PR #36, opened 2021-03-20
Develop

phbasler:
- Removed REQUIRED from boost (should not have been there)
- ThermalFunctions use a static map to save their zeta calculations as they are time expensive
- Reduced the gsl tolerance from 1e-6 to 1e-4. As this corresponds to the allowed error in the global VEV we do not see a big difference in the result, but the runtime is reduced by a large amount
- Added clang style file and the corresponding changes to the code


JonasMueller1991 (approved):
Checked :)

PR #40, opened 2021-05-09
Logger

### LINKED ISSUE #9: Create a Logger function
Instead of the usual information output to std::cout << text we should implement a logger function Log << text which will either write the output to a textfile, to cout or both, depending on a configuration Flag set in utility?

phbasler:
Enabled Logger functionality for all binaries. Use --help to show the available options.

Closes #9


src/utility/Logger.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+
+static std::map<std::string, LoggingLevel> LoggingPrefixes{
+    {
+        "--logginglevel::default=",
+        LoggingLevel::Default,
+    },
+    {"--logginglevel::debug=", LoggingLevel::Debug},
+    {"--logginglevel::ewbgdetailed=", LoggingLevel::EWBGDetailed},
+    {"--logginglevel::progdetailed", LoggingLevel::ProgDetailed},
```
lisabiermann, line 26:
```suggestion
    {"--logginglevel::progdetailed=", LoggingLevel::ProgDetailed},
```

include/BSMPT/utility/Logger.h
```diff
@@ -0,0 +1,126 @@
...
+  void Disable();
+
+  std::ostream mOstream;
+  std::ofstream mfilestream;
+
+  std::map<LoggingLevel, bool> mCurrentSetup{
+      {LoggingLevel::Default, true},
+      {LoggingLevel::EWBGDetailed, false},
+      {LoggingLevel::ProgDetailed, false},
+      {LoggingLevel::ProgDetailed, false},
```
lisabiermann, line 78:
```suggestion
      {LoggingLevel::ProgDetailed, false},
      {LoggingLevel::MinimizerDetailed, false},
```

lisabiermann (changes requested):
Looks good, I'd only have two tiny suggestions.

PR #42, opened 2021-06-13
Ci updates

phbasler:
Adjusted test setup to work correctly for small fluctuations around almost vanishing values.


phbasler:
@jwittbrodt @mb-1054571971 Can you try if ctest runs without error on your local machines on the CI-Updates branch?

Issue #47, opened 2021-09-02
Add cxsm unit tests


phbasler:
Currently there are no unit tests for the cxsm. The same tests as for the other models should be added.

PR #48, opened 2021-09-03
Release of v2.3.1

phbasler:
- CxSM bug fix which caused Re(a1)=Im(a1) = 0 when read from datafile 
- CxSM Example Point
- CxSM Unit Test (Closing #47 )
- Added the check of first and second derivative of CW + CT for the Test executable


phbasler:
Blocked until https://github.com/CMA-ES/libcmaes/pull/226 is accepted

lisabiermann (approved):

PR #88, opened 2022-07-05
Changed unique_ptr to make_unique

### LINKED ISSUE #86: Bug when calling template class
I was contacted by mail that we have a bug when using the template class. 

If we call 

    /build/bin/CalcCT --model=template --input=example/C2HDM_Input.dat --output=build/template_test.tsv --firstline=2 --lastline=2 

then the line https://github.com/phbasler/BSMPT/blob/master/src/models/ClassPotentialOrigin.cpp#L3441 causes a segfault. If we change the --model param to c2hdm, then everything works as expected.

I tried it locally and it is not the rfind part, it is the assignment of the member variable.

phbasler:
Closes #86


phbasler:
@JonasMueller1991 @lisabiermann we can ignore the error on the codecoverage. This seems to be some rounding errors.

lisabiermann (approved):

PR #104, opened 2023-02-10
Deal with negative TanBeta in R2HDM

### LINKED ISSUE #103: Deal with negative TanBeta in R2HDM
We loose the information on the sign of `TanBeta` in the current derivation of `C_CosBeta` and `C_SinBeta` in `ClassPotentialR2HDM.cpp`. 
I propose to add an additional `if`-condition to set the sign of `C_SinBeta` accordingly. This way we also don't loose numerical precision.

lisabiermann:
Added an additional `if` to set sign of `C_SinBeta`. Closes #103.

### DIFF
--- a/src/models/ClassPotentialR2HDM.cpp
+++ b/src/models/ClassPotentialR2HDM.cpp
@@ -228,6 +228,11 @@ void Class_Potential_R2HDM::set_gen(const std::vector<double> &par)
   C_SinBetaSquared = TanBeta * TanBeta * C_CosBetaSquared;
   C_SinBeta        = sqrt(C_SinBetaSquared);
 
+  if (TanBeta < 0) // for beta in 4th quadrant
+  {
+    C_SinBeta *= -1;
+  }
+
   u1 = RealMMix * TanBeta -
        C_vev0 * C_vev0 * C_SinBetaSquared * (L4 + RL5 + L3) / 0.2e1 -
        C_vev0 * C_vev0 * C_CosBetaSquared * L1 / 0.2e1;
--- a/tests/unittests/Test-r2hdm.cpp
+++ b/tests/unittests/Test-r2hdm.cpp
@@ -27,6 +27,16 @@ const std::vector<double> example_point_R2HDM{/* lambda_1 = */ 2.740595,
                                               /* tan(beta) = */ 4.63286,
                                               /* Yukawa Type = */ 1};
 
+const std::vector<double> example_point_R2HDM_negTanBeta{
+    /* lambda_1 = */ 2.740595,
+    /* lambda_2 = */ 0.242356,
+    /* lambda_3 = */ 5.534491,
+    /* lambda_4 = */ -2.585467,
+    /* lambda_5 = */ -2.225991,
+    /* m_{12}^2 = */ -7738.56,
+    /* tan(beta) = */ -4.63286,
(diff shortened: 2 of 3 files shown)


src/models/ClassPotentialR2HDM.cpp
```diff
@@ -228,6 +228,11 @@ void Class_Potential_R2HDM::set_gen(const std::vector<double> &par)
   C_SinBetaSquared = TanBeta * TanBeta * C_CosBetaSquared;
   C_SinBeta        = sqrt(C_SinBetaSquared);
 
+  if (TanBeta < 0) // for beta in 4th quadrant
+  {
+    C_SinBeta *= -1;
+  }
+
```
phbasler, line 235: What happens in the case cos(beta)< 0 ? This would lead to a misassignment of signs if cos(beta) < 0 but sin(beta) > 0

phbasler (changes requested):

phbasler:
@lisabiermann  Can you also increase the version of BSMPT?

src/models/ClassPotentialR2HDM.cpp
lisabiermann: We cannot access the region with cos(beta) < 0, as far as I see it. From TanBeta we only get beta in [-Pi/2, Pi/2] by using [`std::atan`](https://en.cppreference.com/w/cpp/numeric/math/atan). With the above fix, we now also correctly reproduce the region where sin(beta) < 0. The cosine is cos(beta) > 0 for beta in [-Pi/2, Pi/2].

phbasler: Alright. Can you add a unit test for this case?

lisabiermann: I've added a unit test that checks the sign of `C_CosBeta` and `C_SinBeta` in 32fe610 for the `example_point_R2HDM`. Shall I add a second example point with negative `TanBeta` to the check?

lisabiermann: Done in feb36d4.

tests/unittests/Test-r2hdm.cpp
```diff
@@ -82,6 +93,19 @@ TEST_CASE("Checking EWPT for R2HDM", "[r2hdm]")
...
 
+TEST_CASE("Checking sign of Sin and Cos of beta", "[r2hdm]")
+{
+  using namespace BSMPT;
+  using namespace Models;
+  Class_Potential_R2HDM point1, point2;
+  point1.set_gen(example_point_R2HDM);
+  point2.set_gen(example_point_R2HDM_negTanBeta);
+
+  REQUIRE(((point1.C_CosBeta >= 0) and (point2.C_CosBeta >= 0) and
```
phbasler, line 104: Please split this up in four seperate requirements. This way we have an better overview what fails

phbasler (changes requested):

tests/unittests/Test-r2hdm.cpp
lisabiermann: Done in 9723d3b.

lisabiermann:
> @lisabiermann Can you also increase the version of BSMPT?

Version is increased to 2.5.1 in 88806ab.

phbasler (approved):

PR #107, opened 2023-03-19
Feature/sm as input and add unit test for Debye calculations

phbasler:
The goal of this PR is to add a small unit test to check if the calculated Debye corrections matches the expected y_t^2/4. 
For this I had to change how the models get the SM constants. It is now a struct which contains the constants.
The `CalcEtaInterface` and `FChoose` had a small API break. To get the old behaviour the new argument can be set to the function `GetSMConstants()` in which case everything is the same as before.
For unit tests, and maybe others if wanted, can now choose other SM parameters easily.

### DIFF
--- a/tests/unittests/Test-origin.cpp
+++ b/tests/unittests/Test-origin.cpp
@@ -29,8 +29,9 @@ const std::vector<double> example_point_C2HDM{/* lambda_1 = */ 3.29771,
 TEST_CASE("Check f_{abcd}", "[origin]")
 {
   using namespace BSMPT;
+  const auto SMConstants = GetSMConstants();
   std::shared_ptr<BSMPT::Class_Potential_Origin> modelPointer =
-      ModelID::FChoose(ModelID::ModelIDs::C2HDM);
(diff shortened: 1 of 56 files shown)


tests/unittests/Test-origin.cpp
```diff
@@ -24,13 +28,58 @@ const std::vector<double> example_point_C2HDM{/* lambda_1 = */ 3.29771,
...
                                               /* tan(beta) = */ 4.64487,
                                               /* Yukawa Type = */ 1};
+
+} // namespace
+
+/**
+ * @test Check if the automatic Debye corrections match the SM one in the SM
+ * case. This should be y_t^2/4.
+ */
+TEST_CASE("Test Calculate Debye", "[origin-s]")
```
lisabiermann, line 38: Why are you using the new tag `[origin-s]` for the Debye-test? What does it stand for?

phbasler: Fixed it. It was useful during debugging

lisabiermann (approved):

PR #124, opened 2024-04-07
Use conan 2 for setting up

phbasler:
- [x] Use conan2 for setup
- [x] simplify workflows
- [x] Adjusted readme
- [ ] Adjust required jobs (after approval in review)
- [x] Adjust directory in coverage workflow
- [x] Add build missing
- [x] Add build script


lisabiermann:
I would also add some sort of warning that tells the user to use `conan2` (i.e. read the README) when the user tries to `cmake ..` directly without `conan` presets. Is there an elegant way to do this?

profiles/BSMPT/linux-release
```diff
@@ -0,0 +1,9 @@
+[settings]
+arch=x86_64
+build_type=Release
+compiler=gcc
+compiler.cppstd=gnu17
+compiler.libcxx=libstdc++11
```
lisabiermann, line 6: `11`->`17`?

profiles/BSMPT/macos-debug
```diff
@@ -0,0 +1,8 @@
+[settings]
+arch=x86_64
+build_type=Debug
+compiler=apple-clang
+compiler.cppstd=gnu17
+compiler.libcxx=libc++
+compiler.version=13
```
lisabiermann, line 7: Need to increase version here?

profiles/BSMPT/windows-debug
```diff
@@ -0,0 +1,8 @@
+[settings]
+arch=x86_64
+build_type=Debug
+compiler=msvc
+compiler.cppstd=14
```
lisabiermann, line 5: `17`?

profiles/BSMPT/windows-release
```diff
@@ -0,0 +1,8 @@
+[settings]
+arch=x86_64
+build_type=Release
+compiler=msvc
+compiler.cppstd=14
```
lisabiermann, line 5: `17`?

lisabiermann (commented):
If my confusion about the compiler or `libcxx` versions gets cleared, I will approve the PR. Works all fine for me and feels like a very elegant solution.

conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+
+required_conan_version = ">=2.0.0 <3"
+
+
+class BSMPT(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps"
+
+    options = {
+        "enable_tests": [True, False],  # enables the unit tests
```
lisabiermann, line 14: I propose to rename this to `EnableTests` to be consistent with the other names.

profiles/BSMPT/linux-debug
```diff
@@ -0,0 +1,10 @@
+[settings]
+arch=x86_64
+build_type=Debug
+compiler=gcc
+compiler.cppstd=gnu17
+compiler.libcxx=libstdc++11
```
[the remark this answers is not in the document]
phbasler: No, libstdc++17 does not exist. This flag decides about the ABI stability which changed with c++11.

I set compiler.version to what I have on my system. Should we rename it or add more profiles?

profiles/BSMPT/macos-debug
phbasler: I just ran detect on the macos runner as I don't have a mac system.

profiles/BSMPT/linux-debug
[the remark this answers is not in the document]
lisabiermann: Ok, no, then let's leave it as it is, I would say.

profiles/BSMPT/macos-debug
lisabiermann: Looking at the [docs](https://docs.conan.io/2/reference/config_files/profiles.html), `compiler.version=14` seems to exist at least, but there's probably no immediate need for us to increase this. If this is what the macos runner has, that sounds fine.

conanfile.py
lisabiermann, line 14:
```suggestion
        "EnableTests": [True, False],  # enables the unit tests
```

conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+        "MakeAdditionalTesting": [True, False],  # build additional test executables
+        "CompileBaryo": [
+            True,
+            False,
+        ],  # compile the electroweak baryogenesis for the C2HDM
+        "EnableCoverage": [True, False],  # enable code coverage
+        "UseVectorization": [True, False],  # use vectorization for the build
+    }
+    default_options = {
+        "enable_tests": True,
```
lisabiermann, line 29:
```suggestion
        "EnableTests": True,
```

conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+        if self.options.CompileBaryo:
+            self.requires("boost/1.84.0")
+
+        if self.options.UseNLopt:
+            self.requires("nlopt/2.7.1")
+
+    def build_requirements(self):
+        self.tool_requires("cmake/3.29.0")
+
+        if self.options.enable_tests:
```
lisabiermann, line 52:
```suggestion
        if self.options.EnableTests:
```

conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+            apt = Apt(self)
+            apt.install(["lcov"], update=True, check=True)
+
+    def layout(self):
+        cmake_layout(self)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+
+        tc.variables["BUILD_TESTING"] = self.options.enable_tests
```
lisabiermann, line 67:
```suggestion
        tc.variables["BUILD_TESTING"] = self.options.EnableTests
```

README.md
```diff
@@ -70,37 +70,52 @@ If you use this program for your work, please cite
...
-### build
-With the dependencies and options you can build the programm with
-  
-        mkdir build && cd build  
-        cmake (OPTIONS from Dependencies) ..  
-        cmake --build . -j  
-        cmake --build . -j -t doc
+3. [libcmaes](https://github.com/CMA-ES/libcmaes): Additionally to GSL you should either use libcmaes or NLopt. For more details on the libcmaes installation, e.g. possible dependencies, visit their [wiki](https://github.com/CMA-ES/libcmaes/wiki). If you don't want to install or use it, you can set `--options UseLibCMAES=False` when using the detailed build, as described below.
+4. [NLopt](https://nlopt.readthedocs.io/en/latest/): If you do not want to use NLopt, you can set `--options UseNLopt=False` when using the detailed build, as described below.
+5. [Boost](https://www.boost.org/) : This is optional and only required for the Baryogenesis calculations. In order to compile the Baryogenesis calculation, set `--options BSMPTCompileBaryo=True` when using the detailed build, as described below.
```
lisabiermann, line 91:
```suggestion
5. [Boost](https://www.boost.org/) : This is optional and only required for the Baryogenesis calculations. In order to compile the Baryogenesis calculation, set `--options CompileBaryo=True` when using the detailed build, as described below.
```

PR #128, opened 2024-04-24
Add  ubuntu-20.04 to test.yml

phbasler:


.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   tests-fullSetup:
     strategy:
       matrix:
-        setup: [ {os: ubuntu-latest, profile: conan-linux-x86_64-release}, {os: windows-latest, profile: conan-windows-x86_64-release}, {os: macos-latest, profile: conan-macos-x86_64-release}] 
+        setup: [ {os: ubuntu-latest, profile: conan-linux-x86_64-release}, {os: ubuntu-20.04, profile: conan-linux-x86_64-release}, {os: windows-latest, profile: conan-windows-x86_64-release}, {os: macos-latest, profile: conan-macos-armv8-release}]
```
phbasler, line 23: Should we include the ubuntu 20 tests? The packages are missing and because of that it takes a while (extra 7 minutes)

vollous (approved):

.github/workflows/test.yml
vollous: Sure. Always nice to support older systems (clusterssometimes have older compilers)

vollous (approved):

PR #133, opened 2024-04-26
remove static derivatives

phbasler:


phbasler:
@vollous  Any idea why the `Checking phase tracking for BP2`  test now fails? From what I have seen this should not change the results. Both implementations of Nabla were identical and the Hessian ones seemed to differ only in the symmetric elements

phbasler:
So locally it works again with now both `HessianNumerical`.

If I use the `HessianNumerical_BI` (the one before in the BonuceAction) in the `minimum_tracer`, then those BP tests fail.
Do the different numerical algorithms have different conditions?

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master, develop ]
+    branches: [ master, develop, BSMPTv3 ]
```
phbasler, line 11:
```suggestion
    branches: [ master, develop ]
```

src/utility/NumericalDerivatives.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+      xp = phi; // F(x-h, y-h)
+      xp[i] -= eps;
+      xp[j] -= eps;
+      r += V(xp);
+
+      result[i][j] = r / (4 * eps * eps);
+      result[j][i] = r / (4 * eps * eps);
+
+      if (i == j) eps *= 2;
```
phbasler, line 69: I get the idea of halving the epsilon because you would add 2 epsilon to x_i otherwise.
But does the `epsilon /=2` here not remove the factor 4? 
So should this be turned in a different order?
```suggestion
      if (i == j) eps *= 2;
      result[i][j] = r / (4 * eps * eps);
      result[j][i] = r / (4 * eps * eps);


```

vollous: I think the order is correct. We should not change step size in the middle of the calculation.
From [Wikipedia](https://en.wikipedia.org/wiki/Finite_difference#Rules_for_calculus_of_finite_difference_operators) the expression for the double $x$ derivative is

$f_{x x}(x,y)\approx{\frac{f(x+h,y)-2f(x,y)+f(x-h,y)}{h^{2}}}$

and for the $x,y$ derivative is

$f_{x y}(x,y)\approx{\frac{f(x+h,y+k)-f(x+h,y-k)-f(x-h,y+k)+f(x-h,y-k)}{4h k}}$.

If $x=y$ and I don't half the step size then I will calculate a derivative with twice the step size. (By halving it indeed removes the $4$ and matches the expression above).

I have two implementations for the Hessian (which should be corrected) but the difference is only on the diagonal.

phbasler: Ok but then something else is causing the tests to fail if I use the BI Hessian in the minimum_tracer file

phbasler: Just to brainstorm: Could it be that the reference in the unit test is wrong? From what you say the BI method seems to be the correct one but leads to failing unit tests.

vollous: I don't think that unit test is wrong, we just check for the number of phases, and we clearly see some unstable behaviour.
I am not sure why this is happening. **Not** halving the ```eps``` produces stabler results and I actually think this is the correct one since the error is of the same order as the other ones.
I also noticed that increasing the constant that we add to the diagonals of the Hessian matrix in the LocateMinimum produces stabler results, but it does not completely solve the problem. The problem seems to be coming from the Hessian not being very stable, one more reason to implement analytical derivatives.

I think we should use the version that does not halve the step (all tests pass this way) and increase the diagonal constant as a precaution (the tests still pass and computational time is not altered).

phbasler:
@vollous So the checks are green but windows takes approx double the time for the checks. Is this ok for you?

PR #135, opened 2024-04-27
Add check for conan profile

phbasler:
- Add check if the conan profile exist and print error message otherwise
- Add option to use custom profile in setup script


vollous (approved):
Worked on my machine

PR #136, opened 2024-04-27
Reduce diagonal runtime of HessianNumerical

phbasler:
Reduce runtime by simplifying the formula in the loop for diagonal elements and start the loop by running from `j=i+1`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,6 @@ on:
     branches: 
       - master
       - develop
-      - BSMPTv3
   workflow_dispatch:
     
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -6,9 +6,13 @@ name: Unit tests
...
   push:
-    branches: [ master ]
+    branches:
+      - master
   pull_request:
-    branches: [ master, develop ]
+    branches: 
+      - master
+      - develop
+      - BSMPTv3
```
phbasler, line 15:
```suggestion
```

phbasler:
@vollous 
This is a minimal runtime improvement because we avoid a calculation of `V(phi)`  for `i=j` case. If you don't see any problem with it (and the tests run green), we can merge it to the v3 branch

vollous (approved):
Looks good, just formated the code.

PR #140, opened 2024-05-06
Conan use default profile

vollous:
If conan cannot match with any profile shipped with BSMPT v3, we create a new profile equal to the default profile and use that one.


phbasler (commented):
I think we should add something to ensure the c++ version is set to 17 to match our desired version

phbasler (approved):

PR #141, opened 2024-05-06
Fix doxygen start page

phbasler:
- Fix starting page of doxygen docu to main readme
- Exclude other readmes in src as they are only there for repository structure


vollous (approved):
Works on my machine

PR #144, opened 2024-05-15
Added testing for the O(4) symmetric potentials.

vollous:
Added tests that compare O(4) solution with the analytical calculations from [Exact Tunneling Solutions in Multi-Field Potentials (J.R. Espinosa, T. Konstandin)](https://arxiv.org/pdf/2312.12360)


src/bounce_solution/action_calculation.cpp
```diff
@@ -474,7 +474,14 @@ std::vector<double> BounceActionInt::ExactSolutionLin(double l0,
       ss << rho_up << "\t" << l << "\t" << dVdl << "\t" << d2Vdl2 << "\n";
       BSMPT::Logger::Write(BSMPT::LoggingLevel::BounceDetailed, ss.str());
       ss.str(std::string());
-      return (ExactSolutionLin(l0, l / 10., dVdl, d2Vdl2));
+      if (l0 == l0 + (l - l0) / 10.)
```
phbasler, line 477: Isn't this check only true if `l0== 1` ?

src/utility/utility.cpp
```diff
@@ -81,10 +81,52 @@ Transpose(const std::vector<std::vector<double>> &A)
...
 }
 
+double Li2(const double &x)
+{
+  if (x == 0) return 0;
+  if (x == 1) return pow(M_PI, 2) / 6.;
+  if (x < -1) return -pow(M_PI, 2) / 6. - pow(log(-x), 2) / 2. - Li2(1. / x);
+  if (x < 0) return 1 / 2. * Li2(-x * -x) - Li2(-x);
+  if (x > 0.5) return pow(M_PI, 2) / 6. - log(x) * log(1 - x) - Li2(1 - x);
+  double sum = 1e-100;
```
phbasler, line 91:
```suggestion
  double sum = 0;
```

Why not start with 0?

phbasler (changes requested):

phbasler:
Should we increase the bug fix version?

src/utility/utility.cpp
vollous: Prevent a 1/0 on line 94.

src/bounce_solution/action_calculation.cpp
vollous: No, its true if ```l = l0```which is only true if ```(l - l0) / 10.``` is so small compared with ```l0```that gets truncated in the sum.
This checks if the recursive function arrived at the numerical precision limit.

phbasler: Ah, this is an `l` and not a 1.
Maybe it is clearer here to use something like `abs(l-l0) <= eps`?

src/utility/utility.cpp
phbasler: Why not test for it and have a proper case for it?

vollous: I didn't want an additional if statement inside the loop.
I found a middle ground. The sum is initialized at ```sum = x```, which is the first element of the sum at ```k=1```, then the sum loop starts at ```k=2```. I commented on this logic for posterity.

src/bounce_solution/action_calculation.cpp
vollous: Done, I compared the difference with the total length. I also added a check that will prevent most of these issues.

src/bounce_solution/action_calculation.cpp
```diff
@@ -474,7 +474,7 @@ std::vector<double> BounceActionInt::ExactSolutionLin(double l0,
       ss << rho_up << "\t" << l << "\t" << dVdl << "\t" << d2Vdl2 << "\n";
       BSMPT::Logger::Write(BSMPT::LoggingLevel::BounceDetailed, ss.str());
       ss.str(std::string());
-      if (l0 == l0 + (l - l0) / 10.)
+      if (abs((l - l0) / Spline.L) < 1e-10)
```
phbasler, line 477: Isn't this now a different check then before? You have `Spline.L` instead of `l0` now

phbasler (changes requested):

src/utility/utility.cpp
phbasler: What prevents it from becoming zero between the loop?

vollous: By the [properties](https://en.wikipedia.org/wiki/Dilogarithm#Identities) of the dilogarithm, if $x<0$ then recall it recursively with positive arguments. Therefore, in the sum $x>0$ always so
$$\sum_{k=2}^\infty\frac{x^k}{k^2}$$
is a series of positive numbers and it can never vanish.

src/bounce_solution/action_calculation.cpp
vollous: Before I trust truncation so that ```l-l0``` was so small compared to ```l0``` so that it vanished in the sum. 
Now I impose that ```l-l0``` is small compared to ```Spline.L```. I basically used your suggestion of having a defined threshold but added ```Spline.L``` to make it scale invariant.

PR #145, opened 2024-05-15
Conan: Different profiles for different clang versions.

vollous:
Split the macos conan profiles into the different clang versions.

### DIFF
--- a/Setup.py
+++ b/Setup.py
@@ -37,7 +37,7 @@ def get_compiler():
 
     if (sys.platform == "darwin"):
         compiler += "-clang-"
-    compiler += get_gcc_version()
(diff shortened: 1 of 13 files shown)


Setup.py
```diff
@@ -90,12 +94,19 @@ def check_profile(profile):
 
 
 def get_gcc_version():
```
phbasler, line 96: Maybe we should rename this function to something like `get_compiler_version`

phbasler (changes requested):

Setup.py
vollous: Done

phbasler (approved):

PR #149, opened 2024-06-24
Fix logic on CalcTempAtFalseVacFraction()

vollous:
Fixed logic on CalcTempAtFalseVacFraction() where sometimes the precision was not enough

### DIFF
--- a/src/bounce_solution/bounce_solution.cpp
+++ b/src/bounce_solution/bounce_solution.cpp
@@ -754,9 +754,12 @@ double BounceSolution::CalcTempAtFalseVacFraction(const double &false_vac_frac)
     this->SetStoredTemp(T_middle); // update temp storage for inner integral
     double IatT = prefac * Nintegrate_Outer(*this).result;
 
-    bool numerically_unstable = false;
-
-    while (std::abs(T_up / T_down - 1) > 1e-8)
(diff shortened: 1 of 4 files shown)


src/bounce_solution/bounce_solution.cpp
```diff
@@ -754,9 +754,12 @@ double BounceSolution::CalcTempAtFalseVacFraction(const double &false_vac_frac)
     this->SetStoredTemp(T_middle); // update temp storage for inner integral
     double IatT = prefac * Nintegrate_Outer(*this).result;
 
-    bool numerically_unstable = false;
-
-    while (std::abs(T_up / T_down - 1) > 1e-8)
+    while (
+        (std::abs(T_up / T_down - 1) >
+         RelativeTemperatureInCalcTempAtFalseVacFraction * 1e-4) or
```
JonasMueller1991, line 759: Why is there an additional 1e-4? There should be no "magic numbers" if there is an already constant

vollous: You are correct, im being inconsistent. I will fix this. The additional factor is to make sure both conditions are met and we do not get stuck in a loop

phbasler (approved):

PR #151, opened 2024-06-27
Fix transition history for multi-step transitions

lisabiermann:
The temporary storage of the completion temperature used in the check for the transition history was not reset after a found transition. Therefore, only histories for one-step first-order PTs were reported correctly. This PR fixes that.
Thanks to @karoerhardt for pointing this out to us!


vollous (approved):
LGTM

PR #153, opened 2024-07-10
Fix tunneling path standalone

vollous:
Missing "&" that caused pointer to go out of scope. It may have to do with the standard change c++14 -> c++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
-  VERSION 3.0.6
+  VERSION 3.0.7
   LANGUAGES C CXX
   DESCRIPTION
     "BSMPT - Beyond the Standard Model Phase Transitions : A C++ package for the computation of the EWPT in BSM models"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
-  VERSION 3.0.7
+  VERSION 3.0.6
   LANGUAGES C CXX
   DESCRIPTION
     "BSMPT - Beyond the Standard Model Phase Transitions : A C++ package for the computation of the EWPT in BSM models"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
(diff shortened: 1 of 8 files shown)


standalone/TunnelingPath.cpp
```diff
@@ -80,7 +80,7 @@ int main()
     BounceActionInt *ClosestBounceActionInt;
     std::cout << "Found a transitions with Tp =\t"
               << bounce.GetPercolationTemp() << " GeV.\n";
-    for (auto BAInt : bounce.SolutionList)
+    for (auto &BAInt : bounce.SolutionList)
```
phbasler, line 83:
```suggestion
    for (const auto &BAInt : bounce.SolutionList)
```

Your problem is not with the c++17 change. Before you made a copy of this object. If this has a pointer then you make a shallow copy which does not receive updates when it changes from a different thread.

I don't know if it compiles here with const ref, but maybe try it.

vollous: It does not compile.

```bash
[ 98%] Built target GenericTests
/Users/joaofranciscoviana/mega/BSMPT/standalone/TunnelingPath.cpp:88:34: error: assigning to 'BSMPT::BounceActionInt *' from 'BSMPT::BounceActionInt const *' discards qualifiers
        ClosestBounceActionInt = &BAInt;
                                 ^~~~~~
1 error generated.
```

 I mentioned the c++14 because I was getting the correct output at some point and I don't remember changing anything else.

vollous: Fixed. Had to convert some member functions to const to make it work.

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
-  VERSION 3.0.5
+  VERSION 3.0.7
```
phbasler, line 9:
```suggestion
  VERSION 3.0.6
```

README.md
```diff
@@ -4,7 +4,7 @@ SPDX-FileCopyrightText: 2021 Philipp Basler, Margarete Mühlleitner and Jonas M
 SPDX-License-Identifier: GPL-3.0-or-later
 -->
 
-Program: BSMPT version 3.0.5
+Program: BSMPT version 3.0.7
```
phbasler, line 7:
```suggestion
Program: BSMPT version 3.0.6
```

phbasler (changes requested):

phbasler (approved):

PR #158, opened 2024-07-23
Add options Build.py

vollous:
Options can be given to Build.py and they will get passes to Setup.py

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
-  VERSION 3.0.5
+  VERSION 3.0.6
   LANGUAGES C CXX
   DESCRIPTION
     "BSMPT - Beyond the Standard Model Phase Transitions : A C++ package for the computation of the EWPT in BSM models"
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ SPDX-FileCopyrightText: 2021 Philipp Basler, Margarete Mühlleitner and Jonas M
 SPDX-License-Identifier: GPL-3.0-or-later
 -->
 
-Program: BSMPT version 3.0.5
+Program: BSMPT version 3.0.6
```
phbasler, line 7:
```suggestion
Program: BSMPT version 3.0.5
```

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 cmake_minimum_required(VERSION 3.23)
 project(
   BSMPT
-  VERSION 3.0.5
+  VERSION 3.0.6
```
phbasler, line 9:
```suggestion
  VERSION 3.0.5
```

phbasler: I don't think we need to increase the version for this one. There is no change in the actual code but only a utility script

phbasler (changes requested):

README.md
phbasler: I will merge this and adjusted the release version of #153 and will make this one release then.

phbasler (approved):

PR #160, opened 2024-07-26
Fix logic calc temps

vollous:


phbasler (approved):

phbasler:
@vollous I removed the version increase and it will be part of the 3.0.6 release

PR #161, opened 2024-07-29
Fix true vacuum at high temperature

vollous:
## Description

In some models, the BSMPT v2 global minimum does not find the global minimum, just non-global. We added a check to make sure that we always start from the true vacuum.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)

### DIFF
--- a/src/prog/CalcGW.cpp
+++ b/src/prog/CalcGW.cpp
@@ -17,6 +17,7 @@
 #include <BSMPT/utility/parser.h>
 #include <BSMPT/utility/utility.h>
 #include <Eigen/Dense>
+#include <chrono>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
--- a/src/prog/CalcTemps.cpp
+++ b/src/prog/CalcTemps.cpp
@@ -17,6 +17,7 @@
 #include <BSMPT/utility/parser.h>
 #include <BSMPT/utility/utility.h>
 #include <Eigen/Dense>
+#include <chrono>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
--- a/src/prog/MinimaTracer.cpp
+++ b/src/prog/MinimaTracer.cpp
@@ -19,6 +19,7 @@
 #include <BSMPT/utility/parser.h>
 #include <BSMPT/utility/utility.h>
 #include <algorithm> // for copy, max
+#include <chrono>
 #include <fstream>
 #include <iomanip>
 #include <iostream>


src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2401,6 +2401,12 @@ Vacuum::Vacuum(const double &T_lowIn,
       PhasesList[i].id = i;
     }
 
+    // impose that true vacuum at T = Tthigh has id = 0.
+    for (auto &phase : PhasesList)
+      if ((phase.T_high == T_high) and
+          (phase.Get(T_high).potential < PhasesList[0].Get(T_high).potential))
+        std::swap(PhasesList[0], phase);
```
phbasler, line 2408: Are you sure this works? You are swapping an iterator with a reference here.
Maybe use an iterator based for loop and use [iterator_swap](https://en.cppreference.com/w/cpp/algorithm/iter_swap)

phbasler (changes requested):

src/minimum_tracer/minimum_tracer.cpp
phbasler: Or maybe easier:
You could use [min_element](https://en.cppreference.com/w/cpp/algorithm/min_element) with a compare function to get the position of the required element and then use iterator_swap just once.
With this there would only be one swap

vollous: > Are you sure this works? You are swapping an iterator with a reference here. Maybe use an iterator based for loop and use [iterator_swap](https://en.cppreference.com/w/cpp/algorithm/iter_swap)

Change this so that iterators are used. Alsoa added testing for this fix.

src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2416,8 +2415,29 @@ Vacuum::Vacuum(const double &T_lowIn,
...
+                  "No traceable global-minimum phase found at Thigh. Abort.");
+    status_vacuum = StatusTracing::NoMinsAtBoundaries;
+    return;
+  }
+  // Global minimum at Thigh at position 0 of PhasesList
+  for (auto phase = PhasesList.begin(); phase != PhasesList.end(); ++phase)
+    if ((phase->T_high == T_high) and
+        (phase->Get(T_high).potential <
+         PhasesList.begin()->Get(T_high).potential))
+      std::iter_swap(PhasesList.begin(), phase);
```
phbasler, line 2440: Why not use `std::min_element` for this?

src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2416,8 +2415,29 @@ Vacuum::Vacuum(const double &T_lowIn,
...
-  return;
+void Vacuum::EnsureHighTemperatureGlobalMininum()
+{
+  // sort phases in decending T_high
+  std::sort(PhasesList.begin(),
+            PhasesList.end(),
+            [](auto a, auto b) { return a.T_high > b.T_high; });
+
+  // Check that we have a phase that reaches up to T_high
+  if (PhasesList.at(0).T_high < T_high)
```
phbasler, line 2428: This will throw an exception if the list is empty. Can that happen here?

phbasler (changes requested):

src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2416,8 +2415,29 @@ Vacuum::Vacuum(const double &T_lowIn,
...
   }
+}
 
-  return;
+void Vacuum::EnsureHighTemperatureGlobalMininum()
+{
+  // sort phases in decending T_high
+  std::sort(PhasesList.begin(),
+            PhasesList.end(),
+            [](auto a, auto b) { return a.T_high > b.T_high; });
```
phbasler, line 2425: Why not include the potential compare from below here directly in the sorting?

vollous: You can only make the potential comparison if ```phase->T_high == T_high```

phbasler (changes requested):

PR #163, opened 2024-07-31
Match convention with ClassPotentialSM.cpp

lisabiermann:
Changes the potential convention and fermion basis order to match the tensors in `ClassPotentialSM.cpp`.

### DIFF
--- a/include/BSMPT/models/ClassPotentialSM.h
+++ b/include/BSMPT/models/ClassPotentialSM.h
@@ -20,9 +20,46 @@ namespace Models
 
 /**
  * @brief The Class_SM class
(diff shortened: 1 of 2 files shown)


phbasler:
As part of this PR we could also add the documentation to the SM implementation. For comparission  [SM](https://phbasler.github.io/BSMPT/documentation/classBSMPT_1_1Models_1_1Class__SM.html) and [C2HDM](https://phbasler.github.io/BSMPT/documentation/classBSMPT_1_1Models_1_1Class__Potential__C2HDM.html).
This way we have documented which basis we use.

phbasler (approved):

PR #172, opened 2024-11-06
Fix of the sign convention of rotation matrices

### LINKED ISSUE #169: Mismatch in rotation matrix convention leading to sign flips in trilinear couplings
We currently use numerical methods to determine the eigensystem of the rotation matrices. This can lead to a mismatch with the desired rotation matrix parametrization in terms of angles. This mismatch then translates to false sign flips in the derivation of the trilinear couplings in the mass basis calculated by the executable `TripleHiggsCouplingsNLO`.

As all calculations related to phase transitions are done in the interaction basis, not the mass basis, this possible mismatch in the rotation matrix convention does not play a role for the results of the other executables.

cborschensky:
Fixes #169. The mismatch between conventions of rotation matrices is resolved by enforcing the parametrisation of [arXiv:2007.02985](https://arxiv.org/abs/2007.02985), Eqs. (5) ff.

We implement a new function `AdjustRotationMatrix()` which calculates the rotation matrix in the parametrisation of the given paper and ensures that the rotation angles are within $-\frac{\pi}{2}$ and $+\frac{\pi}{2}$. This new rotation matrix, saved in the vector `HiggsRotationMatrixEnsuredConvention`, is then used for the calculation inside `TripleHiggsCouplings()`. The convention for the rotation matrices and angles as calculated by BSMPT now also agrees with the one used in the `ScannerS` code.

Additionally, the indices for the mass eigenstates in the rotation matrix are now computed just once for all implemented models inside the respective `AdjustRotationMatrix()` function (previously, the indices were computed separately in the `write()` and `TripleHiggsCouplings()` functions).

General structure of the changes and implementation for the R2HDM by @lisabiermann.


vollous:
Can you make this PR to the "develop" branch?

cborschensky:
> Can you make this PR to the "develop" branch?

Done!

include/BSMPT/models/ClassPotentialOrigin.h
```diff
@@ -969,6 +974,21 @@ class Class_Potential_Origin
...
   Eigen::MatrixXcd LeptonMassMatrix(const std::vector<double> &v) const;
 
+  /**
+   * Ensures the correct rotation matrix convention
+   */
+  virtual void AdjustRotationMatrix() = 0;
+  /**
+   * Returns true if two values are the same given some relative precision
+   */
+  bool almost_the_same(double a, double b, double rel_precision = 0.01);
```
vollous, line 984: Please move the ```almost_the_same``` functions into ```utility.h```/```utility.cpp```.

src/models/ClassPotentialC2HDM.cpp
```diff
@@ -2430,99 +2349,480 @@ void Class_Potential_C2HDM::TripleHiggsCouplings()
...
+  // check if all other elements of rotation matrix are zero
+  bool zero_element = false;
+  for (std::size_t i = 0; i < NHiggs; i++)
+  {
+    for (std::size_t j = 0; j < NHiggs; j++)
     {
-      NeutralHiggs[i] = HiggsMasses[posN.at(i)];
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_rho1  and (ii == pos_G1 or ii == pos_H1)) or
```
vollous, line 2441: This piece of code is a bit confusing, what is it checking for? Can you refactor it a little bit?

src/models/ClassPotentialC2HDM.cpp
```diff
@@ -2430,99 +2349,480 @@ void Class_Potential_C2HDM::TripleHiggsCouplings()
...
+  // Start with i, j = 1, skip neutral Goldstone
+  for (std::size_t i = 1; i < NHiggs; i++)
+  {
+    for (std::size_t j = 1; j < NHiggs; j++)
     {
-      MhUp   = std::sqrt(NeutralHiggs[1]);
-      MhDown = std::sqrt(NeutralHiggs[0]);
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_si_rho1 and (ii == pos_si_G1 or ii == pos_si_H1)) or
```
vollous, line 2647: What is it checking for? Can you refactor it a little bit?

src/models/ClassPotentialCPintheDark.cpp
```diff
@@ -1340,59 +1287,221 @@ void Class_Potential_CPintheDark::TripleHiggsCouplings()
...
+
+  // check if all other elements of rotation matrix are zero
+  bool zero_element = false;
+  for (std::size_t i = 0; i < NHiggs; i++)
+  {
+    for (std::size_t j = 0; j < NHiggs; j++)
+    {
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_rho1  and ii == pos_Gp) or
```
vollous, line 1366: This piece of code is a bit confusing, what is it checking for? Can you refactor it a little bit?

src/models/ClassPotentialCxSM.cpp
```diff
@@ -784,50 +733,238 @@ void Class_CxSM::TripleHiggsCouplings()
...
+
+  // check if all other elements of rotation matrix are zero
+  bool zero_element = false;
+  for (std::size_t i = 0; i < NHiggs; i++)
+  {
+    for (std::size_t j = 0; j < NHiggs; j++)
+    {
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_Gp    and ii == pos_Gp) or
```
vollous, line 808: This piece of code is a bit confusing, what is it checking for? Can you refactor it a little bit?

src/models/ClassPotentialN2HDM.cpp
```diff
@@ -803,93 +734,301 @@ void Class_Potential_N2HDM::TripleHiggsCouplings()
...
-    testsum = std::abs(HiggsRot(i, 6)) + std::abs(HiggsRot(i, 7)) +
-              std::abs(HiggsRot(i, 8));
-    if (testsum > ZeroThreshold)
+    for (std::size_t j = 0; j < NHiggs; j++)
     {
-      posN.at(countposN) = i;
-      countposN++;
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_rho1  and (ii == pos_G1 or ii == pos_H1)) or
```
vollous, line 833: This piece of code is a bit confusing, what is it checking for? Can you refactor it a little bit?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -959,6 +959,68 @@ Class_Potential_Origin::SecondDerivativeOfEigenvaluesNonRepeated(
...
+                                             double rel_precision)
+{
+  if (std::abs(a) < 1e-10 and std::abs(b) < 1e-10)
+  {
+    return true;
+  }
+  return std::abs(a - b) < std::abs(a + b) / 2 * rel_precision;
+}
+
+bool Class_Potential_Origin::almost_the_same(std::complex<double> a,
```
vollous, line 973: Move this function into ```utility.h```

src/models/ClassPotentialOrigin.cpp
```diff
@@ -959,6 +959,68 @@ Class_Potential_Origin::SecondDerivativeOfEigenvaluesNonRepeated(
...
+  auto inv    = mat.inverse();
+  auto transp = mat.transpose();
+
+  for (std::size_t i = 0; i < NHiggs; i++)
+  {
+    for (std::size_t j = 0; j < NHiggs; j++)
+    {
+      if (!almost_the_same(inv(i, j), transp(i, j), precision))
+      {
+        InvEqTrans = false;
```
vollous, line 1011: Some minor improvements here. If we reach this spot we can return false right away without having to check for all other matrix elements.

src/models/ClassPotentialR2HDM.cpp
```diff
@@ -743,93 +713,270 @@ std::vector<double> Class_Potential_R2HDM::calc_CT() const
...
+  // check if all other elements of rotation matrix are zero
+  bool zero_element = false;
   for (std::size_t i = 0; i < NHiggs; i++)
   {
     for (std::size_t j = 0; j < NHiggs; j++)
     {
-      HiggsRot(i, j) = HiggsRotationMatrix[i][j];
+      int ii = int(i);
+      int jj = int(j);
+      if (not((jj == pos_rho1 and (ii == pos_G1 or ii == pos_H1)) or
```
vollous, line 821: This piece of code is a bit confusing, what is it checking for? Can you refactor it a little bit?

vollous (changes requested):
These changes broke two unit tests:

- 17 - Checking triple higgs NLO couplings in the C2HDM (Failed)
- 43 - Checking triple higgs NLO couplings in the CPINTHEDARK (Failed)

Please check what is happening.

This fix probably comes from some mistake BSMPT was making, can you add a unit test for that specific mistake, so that we prevent it in the future.

lisabiermann:
> These changes broke two unit tests:
> 
>     * 17 - Checking triple higgs NLO couplings in the C2HDM (Failed)
> 
>     * 43 - Checking triple higgs NLO couplings in the CPINTHEDARK (Failed)
> 
> 
> Please check what is happening.
> 
> This fix probably comes from some mistake BSMPT was making, can you add a unit test for that specific mistake, so that we prevent it in the future.

This was expected. We have to re-generate the unit test sources with the corrected trilinear couplings.

include/BSMPT/models/ClassPotentialOrigin.h
cborschensky: I found that the `almost_the_same` function appeared also in `minima_tracer.cpp`/`minima_tracer.h`, so I moved all the versions of the function (with `double`, `std::complex<double>`, and `std::vector<double>` arguments) to `utility.cpp`/`utility.h`.

src/models/ClassPotentialCPintheDark.cpp
```diff
@@ -1340,59 +1287,221 @@ void Class_Potential_CPintheDark::TripleHiggsCouplings()
...
+  }
+
+  // check if all other elements of rotation matrix are zero
+  bool zero_element = false;
+  for (std::size_t i = 0; i < NHiggs; i++)
+  {
+    for (std::size_t j = 0; j < NHiggs; j++)
+    {
+      int ii = int(i);
+      int jj = int(j);
```
phbasler, line 1365: Please use static_cast and not c-style cast.

src/models/ClassPotentialCPintheDark.cpp
```diff
@@ -1340,59 +1287,221 @@ void Class_Potential_CPintheDark::TripleHiggsCouplings()
...
+  // CP in the Dark interaction basis
+  // 0     1     2     3     4      5     6      7     8
+  // rho1, eta1, rho2, eta2, zeta1, psi1, zeta2, psi2, rhoS
+  int pos_rho1 = 0, pos_eta1 = 1, pos_rho2 = 2, pos_eta2 = 3,
+      pos_zeta1 = 4, pos_psi1 = 5, pos_zeta2 = 6, pos_psi2 = 7,
+      pos_rhoS = 8;
+
+  // Indices of mass eigenstates for rotation from interaction to mass basis
+  pos_Gp = -1, pos_Gm = -1, pos_Hp = -1, pos_Hm = -1, pos_HSM = -1,
+  pos_G0 = -1, pos_h1 = -1, pos_h2 = -1, pos_h3 = -1;
```
phbasler, line 1299: It's probably better to use `std::optional<std::size_t>` here. That way you can avoid the casts and still have an unset default value.

phbasler (changes requested):

include/BSMPT/models/ClassPotentialC2HDM.h
```diff
@@ -98,6 +98,10 @@ class Class_Potential_C2HDM : public Class_Potential_Origin
   double CTempC1 = 0, CTempC2 = 0, CTempCS = 0;
   double R_Hh_1 = 0, R_Hh_2 = 0, R_Hh_3 = 0, R_Hl_1 = 0, R_Hl_2 = 0, R_Hl_3 = 0,
          R_Hsm_1 = 0, R_Hsm_2 = 0, R_Hsm_3 = 0;
+
+  int pos_G0, pos_G1, pos_G2, pos_H1, pos_H2, pos_h1, pos_h2, pos_h3;
```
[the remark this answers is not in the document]
lisabiermann: @cborschensky: Could we? (I don't see e.g. `pos_Gp` being calculated anywhere else in that function.)

src/models/ClassPotentialC2HDM.cpp
lisabiermann: This is the sanity check that all entries besides those that were used to identify the basis element identification of the rotation matrix are zero. We loop over all interaction-basis indices and only check those elements whose second index does not equal one of the mass-basis indices that were identified earlier. The comment in line 2654 does not explain that sufficiently?

include/BSMPT/models/ClassPotentialC2HDM.h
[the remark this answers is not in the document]
cborschensky: This has been done in the latest commit.

src/models/ClassPotentialCPintheDark.cpp
cborschensky: Done. I introduced temporary `std::optional` variables, and adapted all data types to `std::size_t` so that the c-style casts mentioned above are not necessary anymore.

src/models/ClassPotentialC2HDM.cpp
lisabiermann: @vollous: Can we mark this as resolved?

src/models/ClassPotentialOrigin.cpp
cborschensky: Fixed. I removed the bool variables and instead immediately return false if any of the `almost_the_same` statements are not fulfilled, and then just return true at the end if it runs through all parts without stopping.

src/utility/utility.cpp
```diff
@@ -115,4 +116,60 @@ double EllipIntSecond(const double &x)
...
+                     const double &num_zero)
+{
+  if (std::abs(a) < num_zero and std::abs(b) < num_zero)
+  {
+    return true;
+  }
+  return std::abs(a - b) < std::abs(a + b) / 2 * rel_precision;
+}
+
+bool almost_the_same(const std::complex<double> &a,
```
phbasler, line 131: This function is not tested by tests.

src/utility/utility.cpp
```diff
@@ -115,4 +116,60 @@ double EllipIntSecond(const double &x)
...
+
+bool almost_the_same(const std::vector<double> &a,
+                     const std::vector<double> &b,
+                     const bool &allow_for_sign_flip,
+                     const double &rel_precision,
+                     const double &num_zero)
+{
+  if (a.size() != b.size())
+  {
+    throw std::runtime_error("Error. Vectors must have the same size.");
```
phbasler, line 149: This case is not tested

src/utility/utility.cpp
```diff
@@ -115,4 +116,60 @@ double EllipIntSecond(const double &x)
...
+  }
+  int count_true = 0;
+  for (std::size_t i = 0; i < a.size(); i++)
+  {
+    if (allow_for_sign_flip)
+    {
+      count_true +=
```
phbasler, line 156: This case is also not tested

phbasler (changes requested):

phbasler:
@cborschensky  @lisabiermann Please have a look at the coverage report. There are a lot of new cases missing a test.

src/utility/utility.cpp
lisabiermann: This function should now be tested for in `Test-gw.cpp`.

src/utility/utility.cpp
lisabiermann: This case should now be tested for in `Test-gw.cpp`.

src/utility/utility.cpp
lisabiermann: Should be now tested for in `Test-gw.cpp`.

src/models/ClassPotentialC2HDM.cpp
```diff
@@ -2396,133 +2328,590 @@ std::vector<double> Class_Potential_C2HDM::calc_CT() const
   return parCT;
 }
 
-void Class_Potential_C2HDM::TripleHiggsCouplings()
+/**
+ * Ensures the correct rotation matrix convention
```
phbasler, line 2332: The documentation should always be in the header file.
In this case the origin class should have it and just pass it through to the overrides, unless they need to specify it.

.github/workflows/linter.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         if: env.LocalCommit == 'true'
 
       - name: Checkout remote branch
-        uses: actions/checkout@v6
+         uses: actions/checkout@v6
```
phbasler:
```suggestion
        uses: actions/checkout@v6
```

phbasler:
@cborschensky Please run the linter locally

cborschensky:
> @cborschensky Please run the linter locally

@phbasler Sorry for asking, but since I have never done this before: do you mean running the linter GitHub action locally (is there an easy way to do this?) or just manually running `clang-format` on the changed files (as is done in `linter.yml`) and then committing the changes?

phbasler (approved):
@vollous From the coding side everything looks fine to me. If you want to check the physics one more time, go for it.

Also please have a look at the code coverage report. If you think there is something vital missing, please add it. Otherwise it is fine for me.

PR #174, opened 2024-11-06
Fix temperature error code despite success status code

### LINKED ISSUE #173: Temperature determined as -1 with success status code
For the temperatures determined via the false vacuum fraction it is possible that a `success` status code is reported but the temperature is `-1`. This is due to different conditions used for the stopping of the while loop and for the return of the temperature in `BounceSolution::CalcTempAtFalseVacFraction`.

Thanks to @cborschensky for bringing this to our attention!

lisabiermann:
## Description

This PR fixes #173.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)


vollous:
Would you mind creating a unit test with Christoph's point so that this never happens again? In case we change the logic in the future.

vollous (dismissed):
Looks good. I added some changes to the test to make it quicker (we were getting timeouts).

src/bounce_solution/bounce_solution.cpp
```diff
@@ -754,14 +754,11 @@ double BounceSolution::CalcTempAtFalseVacFraction(const double &false_vac_frac)
...
-            RelativeTemperatureInCalcTempAtFalseVacFraction *
-                MarginOfCalcTempAtFalseVacFractionBeforeFailure) and
-           (not almost_the_same(
-               int_at_false_vac_frac,
-               IatT,
-               RelativeErrorInCalcTempAtFalseVacFraction *
-                   MarginOfCalcTempAtFalseVacFractionBeforeFailure)))
+    while (not(std::abs(T_up / T_down - 1) <
+                   RelativeTemperatureInCalcTempAtFalseVacFraction and
+               almost_the_same(int_at_false_vac_frac,
```
phbasler, line 759: In the unit test the almost_same part of the while condition is never hit.
Is this just a because the choice of test point or is this in general only a fallback check?

cborschensky:
There were two things regarding the changes in the PR that I noticed:

### Comment 1
The logic of [lines 757-764](https://github.com/phbasler/BSMPT/blob/0aea6b73e6dd2165996b92ad873e54917b8c419c/src/bounce_solution/bounce_solution.cpp#L757) changed. Is this intended? While before, it was of the type `while (A > B and not C)` with
```
A = std::abs(T_up / T_down - 1)
B = RelativeTemperatureInCalcTempAtFalseVacFraction * MarginOfCalcTempAtFalseVacFractionBeforeFailure
C = almost_the_same(int_at_false_vac_frac, IatT, RelativeErrorInCalcTempAtFalseVacFraction * MarginOfCalcTempAtFalseVacFractionBeforeFailure)
```
this PR changes it to `while not (A < B' and C')` with
```
B' = RelativeTemperatureInCalcTempAtFalseVacFraction
C' = almost_the_same(int_at_false_vac_frac, IatT, RelativeErrorInCalcTempAtFalseVacFraction)
```
which is something like `while (A > B' or not C')` when distributing the `not`, so, ignoring the differences between `B` and `B'` as well as `C` and `C'`, there is now an `or` instead of an `and` operator, requiring now both conditions to be fulfilled before the loop exits. I guess this is also related to @phbasler's https://github.com/phbasler/BSMPT/pull/174#discussion_r2013079692.

This change in logic can actually cause an infinite loop, because the integral `IatT` in [line 768](https://github.com/phbasler/BSMPT/blob/0aea6b73e6dd2165996b92ad873e54917b8c419c/src/bounce_solution/bounce_solution.cpp#L768) is not always calculated precisely enough during the bisection, so in such a case, the `almost_the_same(...)` condition is never true! For the example [input_infinite_loop_2.txt](https://github.com/user-attachments/files/19507569/input_infinite_loop_2.txt) with the command line
```
bin/CalcTemps --model=r2hdm --input=input_infinite_loop_2.txt --output=test.tsv --firstline=2 --lastline=2 --logginglevel::bouncedetailed=true
```
I found such a case where the integral `IatT` seems to be very steep, and its value always switches between `0` and a very large value, but never reaches or gets close to the desired value `int_at_false_vac_frac`, and so the infinite loop occurs (the point seems to be a bit unstable, however, as playing around with the settings of the minimizers makes BSMPT exit quickly sometimes, saying the point is not NLO stable).

### Comment 2
The condition of the `while` loop in [lines 757-764](https://github.com/phbasler/BSMPT/blob/0aea6b73e6dd2165996b92ad873e54917b8c419c/src/bounce_solution/bounce_solution.cpp#L757) and the condition in [lines 784-788](https://github.com/phbasler/BSMPT/blob/0aea6b73e6dd2165996b92ad873e54917b8c419c/src/bounce_solution/bounce_solution.cpp#L784) to exit the loop are the same after the changes of the PR. It is thus probably not necessary to keep both?

Maybe, if I may make a suggestion, a safer way to handle this loop could be to do the following:
```
    while (not(std::abs(T_up / T_down - 1) <
                   RelativeTemperatureInCalcTempAtFalseVacFraction or
               almost_the_same(int_at_false_vac_frac,
                               IatT,
                               RelativeErrorInCalcTempAtFalseVacFraction)))
    {
      T_middle = (T_up + T_down) / 2.;
      this->SetStoredTemp(T_middle); // update temp storage for inner integral
      IatT = prefac * Nintegrate_Outer(*this).result;

      Logger::Write(LoggingLevel::BounceDetailed,
                    "Pf ( T = " + std::to_string(T_middle) +
                        " ) = " + std::to_string(std::exp(-IatT)));

      if (IatT < int_at_false_vac_frac)
      {
        T_up = T_middle;
      }
      else
      {
        T_down = T_middle;
      }

      res_Temp = T_middle;
    }
```
i.e. change the `and` in the `while` condition to an `or` (so that the loop exits when at least one of the conditions is fulfilled, and not wait for both), and remove the `if` condition at the end of the loop, just keep the `res_Temp = T_middle;` line. This should not change anything regarding the output, and will make sure that, as soon as the `while` loop is entered, a valid temperature will always be output, and the `while` loop will always exit after a certain number of iterations (when `T_up` and `T_down` are close enough). Unless it is not desired that upon entering the `while` loop, the function `CalcTempAtFalseVacFraction` always returns a valid temperature?

vollous:
Superseded by https://github.com/BSMPT/BSMPT/pull/295

PR #175, opened 2024-11-06
Add mass over temperature ratio check output

lisabiermann:
For each derived characteristic temperature, add one line of output to `LoggingLevel::TransitionDetailed` that informs the user about the thermal mass over temperature ratios at this temperature.


src/transition_tracer/transition_tracer.cpp
```diff
@@ -489,4 +505,53 @@ TransitionTracer::~TransitionTracer()
...
 }
 
-} // namespace BSMPT
\ No newline at end of file
+void TransitionTracer::CheckMassRatio(const user_input &input,
+                                      const std::vector<double> &vec,
+                                      const double &temp) const
+{
+  std::stringstream ss;
+  std::vector<double> massOverTempSq, tmp;
```
vollous, line 513: Maybe change these names ```massOverTempSq, tmp``` to others more meaningful? 
Like ```massOverTempHiggs, massOverTempGauge```?

vollous (commented):
Do you think we should also print the same ratios for the true vacuum? And should we save it in the output file or is just a print sufficient?

lisabiermann:
> Do you think we should also print the same ratios for the true vacuum? And should we save it in the output file or is just a print sufficient?

I'm not sure what this ratio would mean in the true vacuum. My understanding is that it is a measure of how good our high-temperature expansion should work, with the false vacuum as the initial position of the transition, that is why I check the false vacuum thermal masses. 
For the moment, I would leave it as optional print statement. But I'm not against adding it as output in the future.

src/transition_tracer/transition_tracer.cpp
lisabiermann: I renamed `tmp` into `massOverTempSqGauge`, as this is what it is. But kept the `massOverTempSq` because in the next step, we insert `massOverTempSqGauge` in it to have all thermal masses squared in one vector. Better?

vollous: Yes, thank you ^^

vollous (approved):

PR #179, opened 2024-11-27
Implementation of the RxSM model

cborschensky:
The real singlet extension of the Standard Model (RxSM) is implemented as a new model into BSMPT, following the convention of the Lagrangian in https://arxiv.org/abs/1512.05355 Eq. (11).

There are three (or four) input parameters for the model:
- the quartic self-coupling of the singlet `lambdaS`,
- the coupling between doublet and singlet `lambdaHS`,
- the singlet VEV `vS` and, in case of `vS == 0`, additionally the singlet mass `MassS`.

The model is implemented for the case of the symmetric, i.e. unbroken, singlet phase `vS == 0`, and for the case of the broken singlet phase `vS =/= 0`. In the case of the unbroken phase, if the input parameter `vS` is set to zero, the input parameter `MassS` has to be provided. In the case of the broken phase, the input of `MassS` is ignored.

Thus, the input file has to contain four columns: `lambdaS`, `lambdaHS`, `vS`, `MassS`. An example input file is [attached here](https://github.com/user-attachments/files/17935680/rxsm_test_input.txt).


lisabiermann:
@cborschensky: Do you have a good example point for the `rxsm` that we could ship in the `example` folder?

phbasler:
@cborschensky Are the error in the unit tests here the same as in #172 ?
If so we should mark this one as a draft and finish #172 first.
Also could you allow edit by maintainers? I tried something to fix the linter workflow from forks but cant update your branch

PR #180, opened 2024-11-28
BSMPT v3.1.0 pull request

vollous:
### Change log
- Small changes to ```prepareData_*.py``` scripts.
- Changes to the ```python``` model implementation method.
- ```Mathematica``` model implementation method
- Update gravitational wave spectrum https://arxiv.org/abs/2403.03723
- Update sound wave efficiency factor https://arxiv.org/abs/2010.09744
- Calculate d.o.f. $g_*(T)$


phbasler:
We need to adjust the version for the pr

README.md
```diff
@@ -146,11 +146,11 @@ To add a new model, you have to modify/create five files (for further details, a
...
 
 1. At `tools/ModelGeneration/Maple` we provide the maple Worksheet `CreateModel.mw` which you can use to implement your model and get the tensors.
-2. At `tools/ModelGeneration/sympy` we provide a setup using only `python3` with `sympy` (at least version 1.10!, if your packet manager only has an older installed, e.g. ubuntu 20.04 only has v1.6, then you have to install v1.10 or up with pip). Here we provide two examples, `SM.py` and `G2HDM.py` (generic 2HDM) which both use the `ModelGenerator.py` module to calculate the tensors and CT. You can get the CT using `python3 SM.py --show ct` and the tensors by calling `python3 SM.py --show tensors`. If your counterterms don't have a unique solution, then the solution space will be shown to you and you have to add additional equations until you have a unique solution (e.g. in the G2HDM example).
-3. To show the simplified tree-level and counterterm potentials, you can use `python3 SM.py --show treeSimpl` and `python3 SM.py --show CTSimpl`.
+2. At `tools/ModelGeneration/Mathematica` we provide a Mathematica framework to implement your model.
```
phbasler, line 152: Any reason why you made it as number 2? 
We could try to reorder them in a way that the numbering represents the recommend method.

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -81,6 +102,16 @@ class BounceSolution
...
 
+  /**
+   * @brief Temperature at which to calculate parameters
+   * - = 0 - Not set
+   * - = 1 - Approximate Nucleation
+   * - = 2 - Exact Nucleation
+   * - = 3 - Percolation
+   * - = 4 - Completion Temperature
+   */
+  int which_transition_temp = false;
```
phbasler, line 113:
```suggestion
  int which_transition_temp = 0;
```

I find this highly irritating. Why write false here and convert it to 0?
Also why not use an enum class with these values?

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -145,6 +194,30 @@ class BounceSolution
...
   tk::spline S3ofT_spline;
 
+  /**
+   * @brief Gstar spline, T < 214.0
+   *
+   */
+  tk::spline GstarProfileLowT;
+
+  /**
+   * @brief Gstar spline, T > 214.0
```
phbasler, line 204: Where does the cutoff of 214 GeV come from?

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -255,6 +328,19 @@ class BounceSolution
...
+   *
+   * - = 0 - Not set
+   * - = 1 - Approximate Nucleation
+   * - = 2 - Exact Nucleation
+   * - = 3 - Percolation
+   * - = 4 - Completion Temperature
+   *
+   * @param which_transition_temp_in
+   */
+  void SetAndCalculateGWParameters(const int &which_transition_temp_in);
```
phbasler, line 342: Calling this function with an enum makes it more readable

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -324,6 +412,7 @@ class BounceSolution
                  const std::shared_ptr<MinimumTracer> &MinTracer_in,
                  const CoexPhases &phase_pair_in,
                  const double &UserDefined_vwall_in,
+                 const int &UserDefined_PNLO_scaling_in,
```
phbasler, line 415: Why insert the new argument in the middle?
I would recommend inserting it at the end. That way users can just extend there call and not have to reorder their arguments.

Technically you would also need a default value for this (at the end), so that old implementations dont break. Otherwise we have a major break in the API and would need to raise the version to 4.0

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -350,6 +441,7 @@ class BounceSolution
...
```
phbasler, line 444: same as above

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -415,9 +523,25 @@ class BounceSolution
...
+  void InitializedVSpline();
+
+  /**
+   * @brief Get the Gstar object
+   *
+   * @param T temperature
+   * @return double
    */
-  double GetGstar() const;
+  double GetGstar(const double &T) const;
```
phbasler, line 544: Maybe a default value or an overload for old calls?

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -459,10 +583,29 @@ class BounceSolution
...
+  /**
+   * @brief GetReheatingTemp Get reheating temperature
+   */
+  double GetReheatingTemp() const;
+
   /**
    * @brief CalcTransitionTemp Get transition temperature from int
    */
-  double CalcTransitionTemp(const int &which_transition_temp);
+  void CalcTransitionTemp();
```
phbasler, line 599: Why removed the argument?

phbasler (changes requested):

sh/prepareData_R2HDM.py
phbasler, line 1: I think something went wrong with these scripts.
The R2HDM script has two convert functions

sh/prepareData_R2HDM.py
```diff
@@ -15,28 +15,45 @@ def convert(InputFile, OutputFile):
...
+Lambda1 = "L1"
+Lambda2 = "L2"
+Lambda3 = "L3"
+Lambda4 = "L4"
+Lambda5 = "L5"
+tanbeta = "tbeta"
+m12squared = "m12sq"
 
     with open(InputFile, "r") as file:
         df = pd.read_csv(file, index_col=HasIndexCol, sep=Separator)
```
phbasler, line 32: Why is this indented?

sh/prepareData_C2HDM.py
```diff
@@ -6,29 +6,41 @@
...
+    Separator='\t'
+    Type='yuktype'
+    Lambda1='L1'
+    Lambda2='L2'
+    Lambda3='L3'
+    Lambda4='L4'
+    re_Lambda5='re_L5'
+    im_Lambda5='im_L5'
+    tanbeta='tbeta'
+    re_m12squared='re_m12sq'
```
phbasler, line 36: This does not work.
Executing black on this file yields errors.
All variables are defined twice, sometimes with different values.

src/bounce_solution/bounce_solution.cpp
```diff
@@ -472,9 +507,56 @@ void BounceSolution::SetGstar(const double &gstar_in)
...
+  GstarProfileLowT.set_points(TGstarLowT, GstarLowT);
+  GstarProfileHighT.set_boundary(
+      tk::spline::not_a_knot, 0.0, tk::spline::not_a_knot, 0.0);
+  GstarProfileHighT.set_points(TGstarHighT, GstarHighT);
+}
+
+void BounceSolution::InitializedVSpline()
+{
+  std::vector<double> T_list, V_list;
+  for (auto m : phase_pair.false_phase.MinimumPhaseVector)
```
phbasler, line 524:
```suggestion
  for (const auto& m : phase_pair.false_phase.MinimumPhaseVector)
```

src/bounce_solution/bounce_solution.cpp
```diff
@@ -472,9 +507,56 @@ void BounceSolution::SetGstar(const double &gstar_in)
...
+    V_list.push_back(m.potential);
+  }
+  FalsePhaseVSpline.set_boundary(
+      tk::spline::not_a_knot, 0.0, tk::spline::not_a_knot, 0.0);
+  FalsePhaseVSpline.set_points(T_list, V_list);
+
+  T_list.clear();
+  V_list.clear();
+
+  for (auto m : phase_pair.true_phase.MinimumPhaseVector)
```
phbasler, line 536:
```suggestion
  for (const auto& m : phase_pair.true_phase.MinimumPhaseVector)
```

src/bounce_solution/bounce_solution.cpp
```diff
@@ -472,9 +507,56 @@ void BounceSolution::SetGstar(const double &gstar_in)
...
+  {
+    T_list.push_back(m.temp);
+    V_list.push_back(m.potential);
```
phbasler, line 539: Isnt this the same code for false_phase and true_phase? Maybe make a function with the phase as an arugment and call that?

src/bounce_solution/bounce_solution.cpp
```diff
@@ -472,9 +507,56 @@ void BounceSolution::SetGstar(const double &gstar_in)
...
+  TruePhaseVSpline.set_boundary(
+      tk::spline::not_a_knot, 0.0, tk::spline::not_a_knot, 0.0);
+  TruePhaseVSpline.set_points(T_list, V_list);
+}
+
+double BounceSolution::GetGstar(const double &T) const
 {
-  return gstar;
+  const double TinMeV    = T * 1000.;
+  const double TTreshold = TGstarLowT.back(); // = TGstarHighT.front()
```
phbasler, line 549: Why the comment with .front?

src/bounce_solution/bounce_solution.cpp
```diff
@@ -855,6 +963,27 @@ void BounceSolution::CalculateCompletionTemp(const double &false_vac_frac)
...
+  if (alpha < 1)
+  {
+    Treh = GetTransitionTemp(); // no supercooling
+  }
+  else // supercooling
+  {
+    if (vwall < vCJ)
+    {
+      std::stringstream ss;
+      ss << "\n\033[1;93mWARNING: Detected wall velocity outside detonation "
```
phbasler, line 977: What is with the \033[1;93m ?

src/bounce_solution/bounce_solution.cpp
```diff
@@ -1045,13 +1162,49 @@ void BounceSolution::CalculateWallVelocity(const Minimum &false_min,
...
+    std::vector<double> res = modelPointer->MinimizeOrderVEV(minimum.point);
+    return modelPointer->VEff(res, minimum.temp);
+  };
+  const double V_before = V(phase.Get(Tstar + eps));
+  const double V_tstar  = V(phase.Get(Tstar));
+  const double V_after  = V(phase.Get(Tstar - eps));
+  const double dVdT     = (V_before - V_after) / (2. * eps);
+  const double d2VdT2   = (V_before - 2. * V_tstar + V_after) / (eps * eps);
+  const double cs       = sqrt(dVdT / (d2VdT2 * Tstar));
+  if (isnan(cs))
```
phbasler, line 1185: Is this ok for you that this case is not triggered by some tests?

src/gravitational_waves/gw.cpp
```diff
@@ -15,28 +15,55 @@ namespace BSMPT
...
+                  kappa::kappaNuMuModel(pow(data.Csound_true, 2),
+                                        pow(data.Csound_false, 2),
+                                        alpha_eff,
+                                        BACalc.vwall,
+                                        vprofile);
+  // XiShock = fastest fluid shell
+  if (data.kappa_sw > 0)
+    data.XiShock = vprofile.back().front();
+  else
+    data.XiShock = 1.; // Does not matter as efficiency if zero
```
phbasler, line 58:
```suggestion
  if (data.kappa_sw > 0)
    {
      data.XiShock = vprofile.back().front();
    }
  else
    {
      data.XiShock = 1.; // Does not matter as efficiency if zero
     }
```

Just to match the style of the rest of the code

src/gravitational_waves/gw.cpp
```diff
@@ -51,98 +78,184 @@ GravitationalWave::~GravitationalWave()
 {
 }
 
+double GetHtauSH(const double HR, const double K_sw)
```
phbasler, line 81: What should happen for K_sw < 0 as an input?

src/gravitational_waves/gw.cpp
```diff
@@ -51,98 +78,184 @@ GravitationalWave::~GravitationalWave()
...
+}
+
+double GetYpsilon(const double HR, const double K_sw)
+{
+  return 1. - 1. / sqrt(1. + 2. * GetHtauSW(HR, K_sw));
+}
+
 double GravitationalWave::CalcEpsTurb(double epsturb_in)
 {
   if (epsturb_in == -1)
```
phbasler, line 98: This does not match the documentation.
There you say 

f [0..1] set
   * to value, for -1 we use the upper bound from
   * https://arxiv.org/abs/1704.05871

So shouldnt the if check match this?

src/gravitational_waves/gw.cpp
```diff
@@ -220,61 +330,412 @@ Nintegrate_SNR(GravitationalWave &obj, const double fmin, const double fmax)
...
+  {
+    P_NLO = 2 * pow(W_coupling, 2) + pow(Z_coupling, 2) + pow(A_coupling, 2);
+    P_NLO *= pow(Tstar, 4); // pressure NLO normalization
+  }
+  const double alpha_infty    = P_LO / RhoGamma;
+  const double gamma_run_away = Rstar / (3. * R0);
+  if (BACalc.GetPTStrength() - alpha_infty < 0)
+    return 0.; // Not enough pressure to drive the bubble wall.
+  const double gamma_eq = pow((dV - P_LO) / P_NLO, 1. / pnlo_scaling);
+  if (gamma_eq < 1) return 0.; // dV - P_LO < P_NLO. Unphysical gamma < 1;
```
phbasler, line 730: What is with the commented code?
```suggestion
  if (gamma_eq < 1) return 0.; 
```

phbasler (changes requested):

src/transition_tracer/transition_tracer.cpp
```diff
@@ -138,10 +138,14 @@ TransitionTracer::TransitionTracer(user_input &input)
             new_transition_data.crit_false_vev =
                 pair.false_phase.Get(pair.crit_temp).point;
 
+            (void)CheckMassRatio(
+                input, new_transition_data.crit_false_vev, pair.crit_temp);
```
phbasler, line 142: Why do you need these function calls?

tools/ModelGeneration/Mathematica/R2HDM.m
phbasler, line 1: Just as a remark: I don't have mathematica, so I cant review these files properly

tests/unittests/Test-origin.cpp
```diff
@@ -119,3 +121,24 @@ TEST_CASE("Check f_{abcd}", "[origin]")
     REQUIRE(result == Approx(expected).margin(1e-4));
   }
 }
+
+TEST_CASE("Check CheckImplementation", "[origin]")
```
phbasler, line 125: Isn't this the same test as `TEST_CASE("Run CheckImplementation in c2hdm", "[c2hdm]")` in Test-c2hdm? 
I added it with my conan PR, so we probably did the same thing here

tests/unittests/Test-gw.cpp
```diff
@@ -1424,3 +1582,876 @@ TEST_CASE("Espinosa-Konstandin - Three-Field Examples - rho = 1/8", "[gw]")
...
+  REQUIRE(
+      kappaNuMuModel(0.5106836025229592, 0.7773502691896259, 0.0001, 0.35) ==
+      Approx(0.0002001547078822659).epsilon(eps));
+  REQUIRE(
+      kappaNuMuModel(0.5106836025229592, 0.7773502691896259, 0.0001, 0.65) ==
+      Approx(0.0006609907513608744).epsilon(eps));
+  REQUIRE(
+      kappaNuMuModel(0.5106836025229592, 0.7773502691896259, 0.0001, 0.95) ==
+      Approx(0.00048331431171071257).epsilon(eps));
+  REQUIRE(kappaNuMuModel(0.5106836025229592,
```
phbasler, line 2001: You call the function here many times with different paramters, correct?
This looks like a job for parameterised tests https://github.com/catchorg/Catch2/blob/devel/docs/generators.md

phbasler (changes requested):

README.md
lisabiermann: No reason, I reordered them as 1. `python`, 2. `Mathematica`, 3. `Maple` (from open to closed source, most to least widely used).

src/transition_tracer/transition_tracer.cpp
lisabiermann: The `CheckMassRatio` function prints the values of `m^2/T^2`, coloured by how large the ratio is, and returns the maximal ratio. The latter we use in `Test-gw.cpp` to test this function, the former is to give at least some information about the validity of the high-T expansion.

lisabiermann: We call it for each characteristic temperature that we calculate.

src/gravitational_waves/gw.cpp
lisabiermann: This part has changed in the updated manual, we now use the upper bound $\epsilon = \sqrt{1 - \Upsilon}$, see Eq. (3.83) and Eq. (3.77) for the definition of $\Upsilon$.

include/BSMPT/bounce_solution/bounce_solution.h
vollous: It is the QCD phase transition. https://arxiv.org/pdf/1609.04979

include/BSMPT/bounce_solution/bounce_solution.h
lisabiermann: We added an overload for the call with no argument with `CalcGstarPureRad` (effective degrees of freedom in the fully thermalized universe). This way, `GetGstar()` returns the same value as before, and `GetGstar(T)` gives the new temperature-dependent value.

src/gravitational_waves/gw.cpp
phbasler: Then the documentation has to be adjusted. They should match

include/BSMPT/bounce_solution/bounce_solution.h
vollous: Done

include/BSMPT/bounce_solution/bounce_solution.h
vollous: I made this an enum class

tests/unittests/Test-gw.cpp
vollous: But the results were evaluated numerically. Do I save them in a file that gets imported during testing?

src/gravitational_waves/gw.cpp
lisabiermann: What documentation do you refer to?

phbasler: The function documentation in the header file. 

https://github.com/phbasler/BSMPT/blob/18e073d7eb90753a5896c91b7f586b2df5622235/include/BSMPT/gravitational_waves/gw.h#L111-L117

sh/prepareData_C2HDM.py
lisabiermann: I propose to fix the `sh/prepareData_*.py` files by merging #165. Ok?

sh/prepareData_R2HDM.py
lisabiermann: I propose to fix the `sh/prepareData_*.py` files by merging #165. Ok?

sh/prepareData_R2HDM.py
lisabiermann: I propose to fix the `sh/prepareData_*.py` files by merging #165. Ok?

include/BSMPT/bounce_solution/gstar.h
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @file spline to calculate gstar
+ */
+
+#pragma once
+#include <vector>
+
+namespace BSMPT
+{
+extern const std::vector<double> TGstarLowT, TGstarHighT;
```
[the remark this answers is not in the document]
vollous: Added

include/BSMPT/bounce_solution/bounce_solution.h
vollous: No particular reason. Moved to the end and given default value.

include/BSMPT/bounce_solution/bounce_solution.h
vollous: Done same as above

include/BSMPT/gravitational_waves/gw.h
```diff
@@ -15,37 +15,85 @@
...
-  double h2OmegaPeakTurbulence = false; // peak amplitude for turbulence
+  double betaH          = false; // inverse time scale beta/H
+  double vw             = false; // bubble wall velocity
+  double vCJ            = false; // Chapman-Jouguet velocity
+  double XiShock        = false; // Shock speed
+  double Csound_false   = false; // speed sound false vacuum
+  double Csound_true    = false; // speed sound true vacuum
+  double kappa_col      = false; // efficiency factor for collision
+  double kappa_sw       = false; // efficiency factor for sw
+  double K_sw           = false; // kinetic energy fraction for sound waves
```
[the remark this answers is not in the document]
vollous: Better clarity and to match the manual

src/gravitational_waves/gw.cpp
lisabiermann: The documentation is updated now.

src/bounce_solution/bounce_solution.cpp
vollous: Yes, Its just that a ```sqrt(x)```is always dangerous and there is no assurence that the argument $x >0$. This is just to be on the safe side.

src/bounce_solution/bounce_solution.cpp
lisabiermann: This is to make the warning appear in color, for better visibility (like we do with other output in the code).

tests/unittests/Test-origin.cpp
vollous: It is, it was added when we fixed the loop dependency. I will remove this one as it is C2HDM-specific.

tools/ModelGeneration/Mathematica/R2HDM.m
vollous: Just tested it and it is working. We also have another PhD student using it and it works quite well.

lisabiermann: I also tested it for the PR to develop of that branch and know at least one person using it successfully.

src/transition_tracer/transition_tracer.cpp
```diff
@@ -489,4 +533,55 @@ TransitionTracer::~TransitionTracer()
...
+        color = 2;
+        break;
+      }
+    }
+  }
+
+  if (color == 0)
+  {
+    ss << "\n\033[1;92mm^2(vev_false, T = " << std::to_string(temp)
+       << ") / T^2 = " << vec_to_string(massOverTempSq) << "\033[0m\n";
```
phbasler, line 570:
```suggestion
       << ") / T^2 = " << massOverTempSq << "\033[0m\n";
```

src/transition_tracer/transition_tracer.cpp
```diff
@@ -489,4 +533,55 @@ TransitionTracer::~TransitionTracer()
...
+  }
+  else if (color == 1)
+  {
+    ss << "\n\033[1;93mm^2(vev_false, T = " << std::to_string(temp)
+       << ") / T^2 = " << vec_to_string(massOverTempSq) << "\033[0m\n";
```
phbasler, line 575:
```suggestion
       << ") / T^2 = " << massOverTempSq << "\033[0m\n";
```

src/transition_tracer/transition_tracer.cpp
```diff
@@ -489,4 +533,55 @@ TransitionTracer::~TransitionTracer()
...
+  }
+  else
+  {
+    ss << "\n\033[1;91mm^2(vev_false, T = " << std::to_string(temp)
+       << ") / T^2 = " << vec_to_string(massOverTempSq) << "\033[0m\n";
```
phbasler, line 580:
```suggestion
       << ") / T^2 = " << massOverTempSq << "\033[0m\n";
```

include/BSMPT/transition_tracer/transition_tracer.h
```diff
@@ -181,14 +195,24 @@ class TransitionTracer
...
    */
   std::vector<BounceSolution> ListBounceSolution;
 
   /**
    * @brief output data storage
    */
   output output_store;
+
+  /**
+   * @brief CheckMassRatio
```
phbasler, line 207: Can you add a description of the function?

include/BSMPT/gravitational_waves/gw.h
```diff
@@ -171,53 +242,188 @@ struct resultErrorPair
...
+ * @return double
+ */
+double getwow(double a, double b);
+void custom_error_handler(const char *reason,
+                          const char *file,
+                          int line,
+                          int gsl_errno);
+/**
+ * @brief returns the fluid velocity behind the wall, v−, and the expansion mode
+ * (2=detonation, 1=hybrid, 0=deflagration)
```
phbasler, line 323: Wouldn't the second part here also be better as an enum class?

include/BSMPT/gravitational_waves/gw.h
```diff
@@ -171,53 +242,188 @@ struct resultErrorPair
...
+ *
+ * @param al \f$ \alpha \f$
+ * @param vw \f$ v_w \f$
+ * @param cs2b \f$ c_{s,b}^2 $ sound speed squared in the broken phase
+ * @return std::pair<double, int>
+ */
+std::pair<double, int> getvm(double al, double vw, double cs2b);
+/**
+ * @brief encodes the differential equation solved in the shock/rarefaction wave
+ * and returns (dξ/dv, dw/dv).
```
phbasler, line 333: How is the return value an int if it is two derivatives?

src/bounce_solution/bounce_solution.cpp
```diff
@@ -67,7 +69,8 @@ BounceSolution::BounceSolution(
                      MaxPathIntegrations_in,
                      NumberOfInitialScanTemperatures_in,
                      {Eigen::MatrixXd::Identity(pointer_in->get_nVEV(),
-                                                pointer_in->get_nVEV())})
+                                                pointer_in->get_nVEV())},
+                     UserDefined_PNLO_scaling_in)
```
phbasler, line 73: In line 92 you have

`for (auto GroupElement : GroupElements)`

Please change that to `for (const auto& GroupElement : GroupElements)`

src/gravitational_waves/gw.cpp
phbasler: If you are sure of it, then it is fine

src/gravitational_waves/gw.cpp
phbasler: I don't see an update

src/gravitational_waves/gw.cpp
```diff
@@ -220,61 +335,412 @@ Nintegrate_SNR(GravitationalWave &obj, const double fmin, const double fmax)
...
+  std::vector<double> LeptonSq_True =
+      BACalc.modelPointer->LeptonMassesSquared(TrueVacuum);
+  std::vector<double> QuarkSq_True =
+      BACalc.modelPointer->QuarkMassesSquared(TrueVacuum);
+  std::vector<double> GaugeSq_True =
+      BACalc.modelPointer->GaugeMassesSquared(TrueVacuum, Tstar);
+
+  double P_LO = 0; // Pressure at LO. 1 -> 1
+
+  for (auto massSq : HiggsSq_True)
```
phbasler, line 682:
```suggestion
  for (const auto& massSq : HiggsSq_True)
```

Otherwise you are making a lot of copies

src/gravitational_waves/gw.cpp
phbasler: yes, this sounds better

tests/unittests/Test-gw.cpp
phbasler: No, from gtest I know this like 

TEST_P()
{
 REQUIRE(kappaNuMuModel(a,b,c,d) == result;
}

P.Instantiate( { values1, values 2 , ... }

This also has the advantage that each number pair is an extra test and you see better which one fails

phbasler (changes requested):

src/gravitational_waves/gw.cpp
lisabiermann: Hmm.. The commit is [this one](https://github.com/phbasler/BSMPT/pull/180/commits/5eb6961f74698ae2abd883327be63a084d0d5751), I don't know why it's not shown in this conversation.

src/gravitational_waves/gw.cpp
lisabiermann: [Done.](https://github.com/phbasler/BSMPT/pull/180/commits/6e023a438c923d346fb7106393dc9d6b78b8026c)

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -485,6 +624,17 @@ class BounceSolution
...
+  /**
+   * @brief Calculate the false vacuum fraction \f$ I(T)=\frac{4\pi
+   * v_{b}^{3}}{3}\int_{T}^{T_{c}}\frac{\Gamma(T^{\prime})d
+   * T^{\prime}}{T^{\prime4}H(T^{\prime})}\left(\int_{T}^{T^{\prime}}\frac{d\tilde{T}}{H(\tilde{T})}\right)^{3}
+   * \f$
+   *
+   * @param T temperature
+   * @return double
+   */
+  double I(const double &T);
```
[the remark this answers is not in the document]
lisabiermann: The false vacuum fraction is $\mathrm{e}^{-I(T)}$, we derive it in `BounceSolution::CalcFalseVacFraction`. What about `FalseVacFractionExponent` instead?

include/BSMPT/bounce_solution/bounce_solution.h
vollous: Sure, I will add it for this and for the other member var.

src/bounce_solution/bounce_solution.cpp
vollous: Modified this to make it more readable

src/gravitational_waves/gw.cpp
vollous: Updated in new branch

src/transition_tracer/transition_tracer.cpp
vollous: Removed in new branch

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -353,7 +433,8 @@ class BounceSolution
                  const double &UserDefined_epsturb_in,
                  const int &MaxPathIntegrations_in,
                  const size_t &NumberOfInitialScanTemperatures_in,
-                 std::vector<Eigen::MatrixXd> GroupElements_in);
+                 std::vector<Eigen::MatrixXd> GroupElements_in,
```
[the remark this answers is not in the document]
vollous: Corrected in new branch

include/BSMPT/transition_tracer/transition_tracer.h
lisabiermann: [Done.](https://github.com/phbasler/BSMPT/pull/180/commits/ecfc32385f01cf1b0441c08f1bf50ffb81615dd4)

tests/unittests/Test-gw.cpp
vollous: Done. Take a look to check if that was what you wanted.

include/BSMPT/gravitational_waves/gw.h
vollous: Hum, this is the documentation of another function. The `int` here is the detonation mode. This snipped was adapted from another paper.

src/bounce_solution/bounce_solution.cpp
vollous: Done in PR

src/bounce_solution/bounce_solution.cpp
vollous: Sure, added in PR.

include/BSMPT/gravitational_waves/gw.h
vollous: Yes, added in PR

include/BSMPT/gravitational_waves/gw.h
vollous: I make this int instead a enum class, as per your suggestion

src/gravitational_waves/gw.cpp
phbasler: You still write 

`If [0..1] set
   * to value, for -1 we use the upper bound sqrt(1 - Upsilon)`

but the code is 
-1 = sqrt()
else: value

So what should happen to an input value of 2? Or -0.2?

include/BSMPT/bounce_solution/bounce_solution.h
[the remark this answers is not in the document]
phbasler: sounds good.
Otherwise if you think just I(T) is the most readable then we can keep it that way, but I prefer `FalseVacFractionExponent_I`

include/BSMPT/gravitational_waves/gw.h
```diff
@@ -171,53 +242,200 @@ struct resultErrorPair
...
+                          int line,
+                          int gsl_errno);
+/**
+ * @brief returns the fluid velocity behind the wall, v−, and the expansion mode
+ * (2=detonation, 1=hybrid, 0=deflagration)
+ *
+ * @param al \f$ \alpha \f$
+ * @param vw \f$ v_w \f$
+ * @param cs2b \f$ c_{s,b}^2 $ sound speed squared in the broken phase
+ * @return std::pair<double, int>
```
phbasler, line 340:
```suggestion
 * @brief returns the fluid velocity behind the wall, v−, and the expansion mode
 *
 * @param al \f$ \alpha \f$
 * @param vw \f$ v_w \f$
 * @param cs2b \f$ c_{s,b}^2 $ sound speed squared in the broken phase
 * @return std::pair<double, ExpansionMode>
```

src/gravitational_waves/gw.cpp
vollous: Those cases are caught by
```c++
if (UserDefined_epsturb != -1 and
    (UserDefined_epsturb < 0 or UserDefined_epsturb > 1))
{
  Logger::Write(LoggingLevel::Default, "Invalid choice for eps_turb.");
  return false;
}
  ```
in ```CalcGW.cpp```.

include/BSMPT/bounce_solution/bounce_solution.h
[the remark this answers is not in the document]
vollous: Sounds good, fixed in PR.

phbasler (approved):

PR #183, opened 2024-12-03
155 fix dependency loop between models and minimizer

vollous:
Moved ```CheckImplementation``` into ```src/prog/Test.cpp```. Now we the library ```Models``` does not depend on ```Minimizer```.

It seemed like an overkill to create a library just for this.


phbasler:
> > Strictly speaking this is a breaking change as we remove a public function. The idea behind moving it to a new library as a free function was offering this to users as a small change. Moving it purely to Test.cpp removes it from all other custom implementations.
> > We can however accept this possibility and make this new library once someone raises an issue
> 
> Okay, I redid the implementation. For some reason, the checks are not appearing on the PR page but are being done anyway...

It seems the automatic cmake formatter does not trigger the workflows. I will have a look at it

include/BSMPT/models/modeltests/ModelTestfunctions.h
```diff
@@ -63,5 +64,14 @@ TestResults CheckSymmetricTensorQuarks(
 TestResults CheckSymmetricTensorGauge(
     const std::vector<std::vector<std::vector<std::vector<double>>>> &Tensor);
 
+/**
+ * Checks if the tensors are correctly implemented. For this the fermion,
+ * quark and gauge boson masses are calculated and printed next to the values
+ * defined in SMparah.h
+ */
+void CheckImplementation(
+    const std::unique_ptr<Class_Potential_Origin> &modelPointer,
```
phbasler, line 73:
```suggestion
    const Class_Potential_Origin &modelPointer,
```

We dont need it as a pointer in the function. The function calls all the subfunctions as a reference anyway

phbasler (changes requested):
Looks good so far. But maybe we could add a unit test for the missing code coverage?
As this is now a free function we could even use gmock for that

include/BSMPT/models/modeltests/ModelTestfunctions.h
vollous: Done

vollous:
> Looks good so far. But maybe we could add a unit test for the missing code coverage? As this is now a free function we could even use gmock for that

[Free functions are not "mockable" out of the box](http://google.github.io/googletest/gmock_cook_book.html#mocking-free-functions). Should I use an interface or convert everything into an actual class?

phbasler:
No, I meant you could implement a mock of the `Class_Potential_Origin` class and pass this mock to the `CheckImplementation`

Or you can use an existing class like the CH2DM or so

phbasler:
It's basically two different points: 

If we want to ensure that the c2hdm model passes the check implementation function, then we use the c2hdm directly.

If we want to ensure that the CheckImplementation fails when it should fail and pass when it should pass, gmock is an easy option as we can change just one ON_CALL for the test

tests/unittests/Test-origin.cpp
```diff
@@ -119,3 +119,15 @@ TEST_CASE("Check f_{abcd}", "[origin]")
...
+
+TEST_CASE("Check CheckImplementation", "[origin]")
+{
+  using namespace BSMPT;
+  const auto SMConstants = GetSMConstants();
+  std::shared_ptr<BSMPT::Class_Potential_Origin> modelPointer =
+      ModelID::FChoose(ModelID::ModelIDs::C2HDM, SMConstants);
+  modelPointer->initModel(example_point_C2HDM);
+
+  ModelTests::CheckImplementation(*modelPointer,
```
phbasler, line 138: This does not check if the checks passed or failed. I added a given check

phbasler (approved):

PR #185, opened 2024-12-08
Combine cmake and c++ linter in one job

phbasler:


src/minimizer/LibCMAES/MinimizeLibCMAES.cpp
```diff
@@ -26,6 +26,7 @@ namespace BSMPT
 namespace Minimizer
 {
 namespace LibCMAES
+
```
phbasler, line 29:
```suggestion
```

tools/cmake/CodeCoverage.cmake
```diff
@@ -130,6 +130,7 @@ get_property(LANGUAGES GLOBAL PROPERTY ENABLED_LANGUAGES)
 list(GET LANGUAGES 0 LANG)
 
 if("${CMAKE_${LANG}_COMPILER_ID}" MATCHES "(Apple)?[Cc]lang")
+
```
phbasler, line 133:
```suggestion
```

phbasler:
@vollous  This fixes the problem that the auto commit of the linter does not trigger the workflows

vollous (approved):

PR #194, opened 2025-01-17
Update the GW spectrum

vollous:
Spectrum
- https://arxiv.org/abs/2403.03723

Efficiency factor
- https://arxiv.org/abs/2010.09744

Implemented a temperature dependent $g_\star(T)$. We extrapolated it from the [SM curve](https://arxiv.org/pdf/1609.04979#figure.7), scaling it to match the value at $T = \infty$ and $T = 0$.


src/bounce_solution/bounce_solution.cpp
```diff
@@ -948,15 +1049,15 @@ void BounceSolution::CalculatePTStrength()
...
+    // Use recursive method to find solution of
     // alpha = alpha(T_*)
     // T_* =  T_*(alpha, v_wall)
     // v_wall = v_wall(alpha, T_*)
     // Should converge quickly, if fails use default value of v_wall = 0.95
     old_alpha = alpha;
-    CalculatePercolationTemp();
-    Minimum true_min  = phase_pair.true_phase.Get(GetPercolationTemp());
-    Minimum false_min = phase_pair.false_phase.Get(GetPercolationTemp());
+    CalcTransitionTemp(); // Calculation all Ts
```
lisabiermann, line 1058: Is this still needed? `CalculatePTStrength` is called after `CalcTransitionTemp` in `SetAndCalculateGWParameters`, so `GetTransitionTemp` should suffice here, no? Maybe we should drop all explicit calls of `CalcSOMETemp` and only inside this function, if the temperature is `!= -1` execute the `Calc` function once to prevent multiple unnecessary executions. But as `CalcTransitionTemp` is only called these two times, it would also suffice to do the former.

vollous: It is necessary because if $v_w = \{-1, -2\}$ we have to perform a recursive calculation. Because $T_\star$ depends on $v_w$ and $v_w$ depends on $T_\star$. 

I guess we could move ```CalcTransitionTemp()``` down  to reduce one calculation but it is so fast that I did not bother optimizing this.

src/prog/CalcGW.cpp
```diff
@@ -534,6 +552,16 @@ CLIOptions::CLIOptions(const BSMPT::parser &argparser)
     ss << "--trans_string not set, using default value: perc\n";
   }
 
+  try
+  {
+    UserDefined_PNLO_scaling = argparser.get_value<double>("pnlo_scaling");
```
lisabiermann, line 557:
```suggestion
    UserDefined_PNLO_scaling = argparser.get_value<int>("pnlo_scaling");
```
Because it's an `int`.

src/bounce_solution/bounce_solution.cpp
lisabiermann: Oh, I see. Then let's leave it.

include/BSMPT/bounce_solution/hubblerate.h
lisabiermann, line 1: I think naming this `gstar.h` makes it much clearer what is done here. The Hubble rate is calculated in the `BounceSolution` class.

src/bounce_solution/hubblerate.cpp
lisabiermann, line 1: Again, what about naming this `gstar.cpp` instead? Also, could we indicate where the numbers stem from exactly for the sake of clarity?

src/gravitational_waves/gw.cpp
```diff
@@ -220,61 +331,391 @@ Nintegrate_SNR(GravitationalWave &obj, const double fmin, const double fmax)
...
+    P_NLO += GaugeSq_True.at(3) * Z_coupling;
+
+    P_NLO -= GaugeSq_False.at(0) * A_coupling;
+    P_NLO -= 2 * GaugeSq_False.at(1) * W_coupling;
+    P_NLO -= GaugeSq_False.at(3) * Z_coupling;
+  }
+  else if (pnlo_scaling == 2)
+  {
+    P_NLO = 2 * pow(W_coupling, 2) + pow(Z_coupling, 2) + pow(A_coupling, 2);
+    P_NLO *= pow(Tstar, 4); // pressure NLO normalization
```
lisabiermann, line 701:
```suggestion
    P_NLO *= pow(Tstar, 3); // pressure NLO normalization
```
https://arxiv.org/pdf/2412.02645 and https://arxiv.org/pdf/2210.07075 have `Tp^3` here, not `Tp^4`?

src/gravitational_waves/gw.cpp
```diff
@@ -220,61 +331,391 @@ Nintegrate_SNR(GravitationalWave &obj, const double fmin, const double fmax)
...
+  const double alpha_eq       = P_NLO / RhoGamma;
+  const double gamma_run_away = Rstar / (3. * R0);
+  if (BACalc.GetPTStrength() - alpha_infty < 0)
+    return 0.; // Not enough pressure to drive the bubble wall.
+  const double gamma_eq =
+      pow((BACalc.GetPTStrength() - alpha_infty) / alpha_eq, 1. / pnlo_scaling);
+  if (gamma_eq <= 1) return 0.; // dV - P_LO < P_NLO. Unphysical gamma;
+  const double R_eq       = 3. * R0 * gamma_eq / 2.;
+  const double gamma_star = min(gamma_eq, gamma_run_away);
+  const double kappa_col  = (1 - alpha_infty / BACalc.GetPTStrength()) *
```
lisabiermann, line 713: Question: How do we make sure that `kappa_col ~ 0` in the regime where the sound waves dominate, is that encoded in this formula?

lisabiermann (changes requested):
Looks good! I left some minor comments that are quickly resolved. I can take care of the reheating warning.

src/bounce_solution/bounce_solution.cpp
```diff
@@ -855,6 +945,18 @@ void BounceSolution::CalculateCompletionTemp(const double &false_vac_frac)
   return;
 }
 
+void BounceSolution::CalculateReheatingTemp()
```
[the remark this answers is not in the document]
lisabiermann: Done with https://github.com/phbasler/BSMPT/pull/194/commits/47f4965b2aa2f797dd8aa4a18c6fecd91a4ea7ff.

src/gravitational_waves/gw.cpp
lisabiermann: True, you are right, for $\gamma^1$, the scaling is $T_{\*}^{3}$, you were correct. But isn't the $T_*^4$ factor missing in case `pnlo_scaling == 1`?

include/BSMPT/bounce_solution/bounce_solution.h
```diff
@@ -45,11 +45,20 @@ class BounceSolution
    */
   std::shared_ptr<Class_Potential_Origin> modelPointer;
 
+  /** Relativistic dof for the energy at T = 0 */
+  const double neutrinogstar = 3.384;
```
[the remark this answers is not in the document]
vollous: I removed this just to match the values I got from the plot digitizer.

include/BSMPT/bounce_solution/hubblerate.h
vollous: Great idea!

src/bounce_solution/hubblerate.cpp
vollous: Great idea!

src/gravitational_waves/gw.cpp
vollous: I expected the expressions to reflect this. If $\Delta V$ then $\gamma_\text{eq} < 1$ so $\kappa_\mathrm{col} = 0$ and so on.

lisabiermann (approved):
Looking good!

vollous:
> @vollous Is there a reason not to merge this?

Yes, we are still waiting for an answer regarding the implementation. Whether it is correct or not.

src/gravitational_waves/gw.cpp
```diff
@@ -51,98 +78,184 @@ GravitationalWave::~GravitationalWave()
...
-void GravitationalWave::CalcPeakAmplitudeSoundWave()
+double GravitationalWave::CalculateXiShell()
 {
-  double ratio = 0;
-  if (IsFluidTurnoverApproxOne(this->data.HR, this->data.K_sw))
-  {
-    ratio = this->data.HR * std::pow(this->data.K_sw, 2.);
-  }
+  double XiFront, XiRear;
+  if (data.vw > data.vCJ)
```
lisabiermann, line 138:
```suggestion
  if (data.vw >= data.vCJ)
```
.. to match Eq. (3.73) in the paper.

src/bounce_solution/bounce_solution.cpp
```diff
@@ -1098,4 +1245,31 @@ double BounceSolution::GetInvTimeScale()
...
+void BounceSolution::CalculateRstar()
+{
+  gsl_integration_workspace *workspace = gsl_integration_workspace_alloc(1000);
+
+  gsl_function F;
+  F.function = [](double T, void *params) -> double
+  {
+    return static_cast<BounceSolution *>(params)->TunnelingRate(T) *
+           exp(-static_cast<BounceSolution *>(params)->I(T)) /
+           (static_cast<BounceSolution *>(params)->HubbleRate(T) * pow(T, 4));
```
lisabiermann, line 1263:
```suggestion
           (static_cast<BounceSolution *>(params)->HubbleRate(T) * pow(T, 2));
```
.. to match https://arxiv.org/pdf/2007.15586, Eq. (3.7)?

src/bounce_solution/bounce_solution.cpp
```diff
@@ -1098,4 +1245,31 @@ double BounceSolution::GetInvTimeScale()
...
+  };
+  F.params = this; // Pass `this` pointer as parameters
+
+  double result, error;
+  gsl_integration_qags(
+      &F, Tstar, Tc, RelErr, AbsErr, 1000, workspace, &result, &error);
+
+  gsl_integration_workspace_free(workspace);
+
+  this->Rstar = pow(pow(Tstar, 3) * result, -1 / 3.);
```
lisabiermann, line 1273:
```suggestion
  this->Rstar = pow(pow(Tstar, 1) * result, -1 / 3.);
```
.. to match https://arxiv.org/pdf/2007.15586, Eq. (3.7)?

src/gravitational_waves/gw.cpp
```diff
@@ -51,98 +78,184 @@ GravitationalWave::~GravitationalWave()
...
+  return 1. - 1. / sqrt(1. + 2. * GetHtauSW(HR, K_sw));
+}
+
 double GravitationalWave::CalcEpsTurb(double epsturb_in)
 {
   if (epsturb_in == -1)
   {
-    double HtauSW = 2. / std::sqrt(3) * data.HR / std::sqrt(data.K_sw);
-    return std::pow((1 - std::min(HtauSW, 1.)), 2. / 3.);
+    return std::pow((1 - GetYpsilon(data.HR, data.K_sw)), 2. / 3.);
```
lisabiermann, line 100:
```suggestion
    return std::sqrt(1 - GetYpsilon(data.HR, data.K_sw));
```
.. to match our Eq. (3.82).

lisabiermann (changes requested):

src/gravitational_waves/gw.cpp
vollous: Correct!

src/bounce_solution/bounce_solution.cpp
lisabiermann: I take back this comment, the implementation is correct.

src/bounce_solution/bounce_solution.cpp
lisabiermann: Same here, all fine as it is.

PR #214, opened 2025-04-12
Fix cmake generation with cmake v4

### LINKED ISSUE #210: CMake 4.0.0 cannot compile BSMPT

phbasler:
Fix cmake generaton for cmake 4.0.0 by removing unsupported version check
Closes #210


vollous (approved):
Works for me

PR #216, opened 2025-04-14
Check if executables were created

vollous:


vollous:
Skipped tests count as failed.

.github/workflows/conan.yml
```diff
@@ -40,3 +40,32 @@ jobs:
           - run: python3 Setup.py --build-missing
 
           - run: python3 Setup.py --create --build-missing --options EnableTests=True
+
```
phbasler, line 43: The conan workflow does not create the executables.
This is by design. You dont need the executables if you want to use the library

vollous: This was the point. It tests if they were not created and complains if they did.

PR #219, opened 2025-04-15
Add unit test to see if BSMPT executable was built into bin(/*)

phbasler:


tests/unittests/CheckExecutables.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <vector>
+
+namespace fs = std::filesystem;
+
+TEST_CASE("Check for existence of BSMPT executable", "[executable]")
+{
+
+  std::vector<fs::path> matching_files;
+
+  std::cout << "Start test" << std::endl;
```
vollous, line 15: Is it ok to use `std::cout <<` here?

phbasler: Yes,

In case that the test fails this is shown in the terminal

vollous (approved):

PR #228, opened 2025-04-30
Use -dumpversion flag to get gcc/clang version

vollous:
## Description

This patch uses the `-dumpversion` flag of compilers to get the compiler version.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)

### DIFF
--- a/Setup.py
+++ b/Setup.py
@@ -121,7 +121,7 @@ def get_compiler_version(compiler: Compiler):
     if sys.platform == "linux" and compiler != Compiler.clang:
         semver_string = subprocess.check_output(
             "gcc -dumpversion".split(), encoding="UTF-8"
-        )#.split("\n")[0]
+        ).split("\n")[0]
         print("call response|", semver_string, "|end")
         return semver_string.partition(".")[0]
     elif sys.platform == "darwin" or compiler == Compiler.clang:
--- a/Setup.py
+++ b/Setup.py
@@ -120,7 +120,6 @@ def get_compiler_version(compiler: Compiler):
         semver_string = subprocess.check_output(
             "gcc -dumpversion".split(), encoding="UTF-8"
         ).split("\n")[0]
-        print("call response|", semver_string, "|end")
         return semver_string.partition(".")[0]
     elif sys.platform == "darwin" or compiler == Compiler.clang:
         semver_string = subprocess.check_output(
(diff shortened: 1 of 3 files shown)


phbasler:
I think this should go directly to main.
This fixes a break for some system and so others should not have to wait for the next dev doployment

Setup.py
```diff
@@ -118,13 +118,13 @@ def check_profile(profile):
...
+            "gcc -dumpversion".split(), encoding="UTF-8"
         ).partition("\n")[0]
         semver_string = version_response[version_response.rfind(" ") + 1 :]
         return semver_string.partition(".")[0]
     elif sys.platform == "darwin" or compiler == Compiler.clang:
         version_response = subprocess.check_output(
-            "clang --version".split(), encoding="UTF-8"
+            "clang -dumpversion".split(), encoding="UTF-8"
         ).partition("\n")[0]
         semver_string = version_response[version_response.rfind("version") + 8 :]
```
cborschensky, line 129: When using the `-dumpversion` flag, this line will cause the output to break for clang, since there is no "version" string in the output anymore. The list slicing should not be necessary anymore here, I think, and you can set `semver_string = version_response` right away.

Setup.py
```diff
@@ -118,13 +118,13 @@ def check_profile(profile):
 def get_compiler_version(compiler: Compiler):
     if sys.platform == "linux" and compiler != Compiler.clang:
         version_response = subprocess.check_output(
-            "gcc --version".split(), encoding="UTF-8"
+            "gcc -dumpversion".split(), encoding="UTF-8"
         ).partition("\n")[0]
         semver_string = version_response[version_response.rfind(" ") + 1 :]
```
cborschensky, line 123: With `-dumpversion`, there will not be a space in the output anymore. At least for me, however, this line does not break the output (unlike the line below for clang). Still, the list slicing should not be necessary here, so `semver_string = version_response` should work also.

vollous: Thanks a lot, fixed.

Setup.py
vollous: Fixed

Setup.py
vollous: Actually, github's gcc outputs
```bash
13\n
```
so we need the "\n" split

vollous:
@phbasler ready for review

phbasler (approved):

PR #237, opened 2025-05-26
Bump

vollous:


.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,150 @@
+name: Bump version
+
+on: 
+  pull_request:
+      branches: [ master ]
+      types: [opened, edited, reopened]
```
phbasler, line 6:
```suggestion
      types: 
        - opened
        - edited
        - reopened
        - synchronize
      
```

We could even think about not triggering this workflow on a draft. Then we would need to add `ready_for_review` to the list

.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,150 @@
...
+              echo "bumpmode=minor" >> $GITHUB_OUTPUT
+            fi
+            if [[ $((patch)) -eq 1 ]]; then 
+              echo "bumpmode=patch" >> $GITHUB_OUTPUT
+            fi
+            if [[ $((nobump)) -eq 1 ]]; then 
+              echo "bumpmode=nobump" >> $GITHUB_OUTPUT
+            fi
+          else
+            echo "Invalid version bump selected."
```
phbasler, line 53:
```suggestion
            echo "You need to choose exactly one checkbox for versioning." >> $GITHUB_STEP_SUMMARY
```

phbasler (changes requested):

lisabiermann (approved):

PR #242, opened 2025-06-24
205 infinite loops and potential memory leaks when running calctempscalcgw

vollous:
## Description

- Fix temperature infinite loops.
- Give ID to patched phase.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)

### DIFF
--- a/src/minimum_tracer/minimum_tracer.cpp
+++ b/src/minimum_tracer/minimum_tracer.cpp
@@ -2286,15 +2286,17 @@ Vacuum::Vacuum(const double &T_lowIn,
                const double &T_highIn,
                std::shared_ptr<MinimumTracer> &MinTracerIn,
                std::shared_ptr<Class_Potential_Origin> &modelPointerIn,
-               const int &UseMultiStepPTMode,
+               const int &UseMultiStepPTModeIn,
                const int &num_pointsIn,
-               const bool &do_only_tracing)
+               const bool &do_only_tracingIn)
 {
-  T_low        = T_lowIn;
-  T_high       = T_highIn;
-  MinTracer    = MinTracerIn;
-  modelPointer = modelPointerIn;
-  num_points   = num_pointsIn;
+  T_low              = T_lowIn;
+  T_high             = T_highIn;
+  MinTracer          = MinTracerIn;
+  modelPointer       = modelPointerIn;
+  num_points         = num_pointsIn;
+  UseMultiStepPTMode = UseMultiStepPTModeIn;
+  do_only_tracing    = do_only_tracingIn;
 
   status_vacuum =
       StatusTracing::Success; // flipped to error code if error encountered
@@ -2407,31 +2409,7 @@ Vacuum::Vacuum(const double &T_lowIn,
        status_vacuum == StatusTracing::NoCoverage)) // no_coverage can get fixed
                                                     // in setCoexRegion
   {
-    // sort phases in decending T_high
-    std::sort(PhasesList.begin(),
-              PhasesList.end(),
-              [](auto a, auto b) { return a.T_high > b.T_high; });
-
-    // assign ids to phases
-    for (std::size_t i = 0; i < PhasesList.size(); i++)
-    {
-      PhasesList[i].id = i;
-    }
-
-    // identify coexisiting phase regions
-    setCoexRegion(UseMultiStepPTMode); // can flip status_vacuum to error code
-
-    if (PhasesList.size() > 0)
-    {
-      PrintPhasesDiagram();
-    }
-
-    if ((status_coex_pairs == StatusCoexPair::Success) and
-        (not do_only_tracing))
-    {
-      // identify coexisting phase pairs
-      setCoexPhases();
-    }
+    orderPhases();
   }
 
   return;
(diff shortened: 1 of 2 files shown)


src/prog/CalcGW.cpp
```diff
@@ -469,15 +464,23 @@ CLIOptions::CLIOptions(const BSMPT::parser &argparser)
...
     {
-      UseMultiStepPTMode = std::stoi(multistepPT_string);
+      int intUseMultiStepPTMode = std::stoi(multistepPT_string);
+      switch (intUseMultiStepPTMode)
+      {
+      case 0: UseMultiStepPTMode = MultiStepPTMode::OneStep; break;
+      case 1: UseMultiStepPTMode = MultiStepPTMode::EdgeCoverage; break;
+      case 2: UseMultiStepPTMode = MultiStepPTMode::CompleteCoverage; break;
+
+      default: ss << "--multistepmode invalid using 'default' instead"; break;
```
phbasler:
```suggestion
      default: 
         { 
            ss << "--multistepmode invalid using 'default' instead"; 
            UseMultiStepPTMode = MultiStepPTMode:Default;
            break;
           }
```

I would set it explicit here, just to be sure

src/prog/CalcTemps.cpp
```diff
@@ -433,15 +428,23 @@ CLIOptions::CLIOptions(const BSMPT::parser &argparser)
...
     {
-      UseMultiStepPTMode = std::stoi(multistepPT_string);
+      int intUseMultiStepPTMode = std::stoi(multistepPT_string);
+      switch (intUseMultiStepPTMode)
+      {
+      case 0: UseMultiStepPTMode = MultiStepPTMode::OneStep; break;
+      case 1: UseMultiStepPTMode = MultiStepPTMode::EdgeCoverage; break;
+      case 2: UseMultiStepPTMode = MultiStepPTMode::CompleteCoverage; break;
+
+      default: ss << "--multistepmode invalid using default instead"; break;
```
phbasler: Here also set default

src/prog/MinimaTracer.cpp
```diff
@@ -422,15 +417,23 @@ CLIOptions::CLIOptions(const BSMPT::parser &argparser)
...
     {
-      UseMultiStepPTMode = std::stoi(multistepPT_string);
+      int intUseMultiStepPTMode = std::stoi(multistepPT_string);
+      switch (intUseMultiStepPTMode)
+      {
+      case 0: UseMultiStepPTMode = MultiStepPTMode::OneStep; break;
+      case 1: UseMultiStepPTMode = MultiStepPTMode::EdgeCoverage; break;
+      case 2: UseMultiStepPTMode = MultiStepPTMode::CompleteCoverage; break;
+
+      default: ss << "--multistepmode invalid using default instead"; break;
```
phbasler: Here also set default

phbasler (changes requested):

phbasler:
@vollous Please have a look a the missing coverage and at the numeric compare remarks from copilot.

src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2561,120 +2574,122 @@
...
 
-    int EdgeOfPhaseatTemp = 0;
-    for (auto temp : tempList)
+  int EdgeOfPhaseatTemp = 0;
+  for (auto temp : tempList)
+  {
+    for (auto edge : edgesList)
     {
-      for (auto edge : edgesList)
+      if (edge.temp == temp)
```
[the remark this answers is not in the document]
vollous: This would breake the logic, we want exact equalities.

src/minimum_tracer/minimum_tracer.cpp
```diff
@@ -2501,8 +2526,9 @@
     Phase phase(Temp, T_high, T_low, MinTracer);
     addPhase(phase);
     print(phase);
-
-    MultiStepPTTracer(phase.T_low, -1);
+    // phase found and we managed to go lower in temperature
+    if (phase.MinimumPhaseVector.size() > 1 and phase.T_low != Temp)
```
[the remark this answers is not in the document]
vollous: This prevents infinite loops

vollous:
> @vollous Please have a look a the missing coverage and at the numeric compare remarks from copilot.

The copilot fixes are not relevant, in fact they would break some logic. The coverage I am not sure how to improve, those are fallback for edge cases.

vollous:
Let's see if this is enough for those edge cases.

vollous:
> > Let's see if this is enough for those edge cases.
> 
> You are missing a test to trigger the `crit_status = BSMPT::StatusCrit::CoincideSinglePoint` case in the tracer, which I think was the point of this PR? Simple test to check if this returns the enum you expect to:
> 
>     * Setup an Instance of `CoexPhases` with T_high == T_low.
> 
>     * Check the crit_status
> 
> 
> As this does not calculate anything this test is instant and you are secured that this returns the wanted enum

I've added this and it slightly improved coverage but it seems that some tests (that locally at least add additional coverage) and not improving it... Particularly this https://github.com/phbasler/BSMPT/blob/7430b8c7399006137309f3b56014a28ffe16b543/src/minimum_tracer/minimum_tracer.cpp#L2648-L2673

phbasler (approved):

PR #246, opened 2025-07-07
Fix gsl warning integration limits and/or step direction not consistent at odeinitval2

vollous:
## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [ ] Bug fix (patch)


phbasler:
Do we need tests for that?
Or maybe some comments why this is necessary?

vollous:
> Do we need tests for that? Or maybe some comments why this is necessary?

Added a test for this scenario.

phbasler (approved):

PR #247, opened 2025-07-20
Update cmaes to version 0.10.2

phbasler:
## Description

Remove the self made cmaes package and use the official cmaes conan package.
Updates the dependency from 0.10.0 + patch to 0.10.2 tag.

Also update the windows conan profiles, otherwise boost would not build anymore with the latest Visual Studio update.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 cmake_minimum_required(VERSION 3.23)
-set(BSMPT_VERSION 3.1.3)
+set(BSMPT_VERSION 3.1.4)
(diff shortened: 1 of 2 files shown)


conanfile.py
```diff
@@ -68,7 +68,7 @@ def requirements(self):
 
         if self.options.UseLibCMAES:
             self.requires(
-                "cmaes/0.10.0@bsmpt/local",
+                "cmaes/0.10.2@bsmpt/local",
```
phbasler, line 71:
```suggestion
                "cmaes/0.10.2",
```

PR #283, opened 2026-02-28
Implementation of the RxSM model

cborschensky:
## Description

The real singlet extension of the Standard Model (RxSM) with a Z2 symmetry is implemented as a new model into BSMPT, following the convention of the Lagrangian in https://arxiv.org/abs/1512.05355 Eq. (11).

The model is implemented both for the case of the unbroken Z2 symmetry (singlet VEV `vS == 0`) and for the case of the spontaneously broken Z2 symmetry (`vS =/= 0`). For the broken phase, the three input parameters are:
- the quartic self-coupling of the singlet `lambdaS`,
- the coupling between doublet and singlet fields `lambdaHS`,
- the singlet VEV `vS`.

For the symmetric phase, the third parameter `vS` must be set to zero. Then a fourth column needs to be added, denoting the mass of the singlet field `MassS`. In the case of the broken phase, the fourth column is ignored.

Thus, the input file has to contain three or four columns: `lambdaS`, `lambdaHS`, `vS`, `MassS`. An example input file both for the case of the broken and unbroken phase is provided in the `example` directory.

## Type of Change

 - [ ] Break (major)
 - [x] Feature (minor)
 - [ ] Bug fix (patch)


example/RxSM_noVS_Input.dat
vollous: Isnt the $v_s = 0$ case included in the other `RxSM` implementation?

tests/GenerateTestCompares/CMakeLists.txt
```diff
@@ -3,9 +3,25 @@
...
 
-set(src C2HDM.cpp R2HDM.cpp N2HDM.cpp CPINTHEDARK.cpp CXSM.cpp SM.cpp)
+set(src
+    C2HDM.cpp
+    R2HDM.cpp
+    N2HDM.cpp
+    CPINTHEDARK.cpp
+    CXSM.cpp
+    RXSM.cpp
+    RXSM_noVS.cpp
```
vollous: Isnt the $v_s = 0$ case included in the other `RxSM` implementation?

include/BSMPT/models/ClassPotentialRxSM.h
```diff
@@ -0,0 +1,101 @@
...
+
+namespace BSMPT
+{
+namespace Models
+{
+
+/**
+ * @brief The Class_Potential_RxSM class
+ * Lagrangian from https://arxiv.org/pdf/1512.05355 Eq. (11)
+ */
```
vollous: Please add more documentation here. Similar to the [CxSM](https://github.com/BSMPT/BSMPT/blob/e9262402097d7a8e36acc6bbb05deeb45db801fb/include/BSMPT/models/ClassPotentialCxSM.h#L21-L65).

vollous (changes requested):
Maybe I missed something but it seems you left some references for the $v_s = 0$ case

vollous (approved):

PR #299, opened 2026-05-08
Boto, Branchina, Borschensky: corrections to T and scalar derivatives…

### LINKED ISSUE #130: Implement analytic derivatives
The implementation of analytical derivatives would improve BSMPT's speed regarding the minimum tracker and the bounce solver. This should be done in the future.

rafaelboto:
## Description

- Fix the analytical gradient of the potential with respect to the fields
- Use $J_{B/F}$ interpolated for negative values on the gradients
- Fix $c_i$ of the Coleman-Weingerg for the Gauge bosons (Parwani)
- `Class_Potential_Origin::boson` also takes `DevMassSquared` (mass squared field derivative)
- Add tests to compare the analytical grad and numerical grad.

classpotential origin 

## Type of Change

 - [ ] Break (major)
 - [x] Feature (minor)
 - [ ] Bug fix (patch)


rafaelboto:
also line 3283 NHiggs should be NGauge

src/models/ClassPotentialOrigin.cpp
```diff
@@ -3248,19 +3268,25 @@ double Class_Potential_Origin::V1Loop(const std::vector<double> &v,
...
+        if (GaugeMassesVec[k] > 0) {
+          VDebye += std::pow(GaugeMassesVec[k], 1.5);
+          VDebye += 1.5*Temp*GaugeMassesVec[k + NGauge]*std::sqrt(std::abs(GaugeMassesVec[k])); 
+        }
         if (GaugeMassesZeroTempVec[k] > 0)
           VDebye += -std::pow(GaugeMassesZeroTempVec[k], 1.5);
       }
 
       VDebye *= -1.0 / (12 * M_PI);
-      res += VDebye;
```
cborschensky, line 3273: I think this line was removed by accident. It should be there, otherwise the whole `VDebye` part is not used in the result of the temperature derivative.

rafaelboto: Added back

src/models/ClassPotentialOrigin.cpp
```diff
@@ -113,30 +113,50 @@ double Class_Potential_Origin::boson(double MassSquared,
...
+           ThermalFunctions::JbosonInterpolated(Ratio, 1);
   }
   else if (diff == -1)
   {
     res += 1.0 / (2 * std::pow(M_PI, 2)) *
            (4 * std::pow(Temp, 3) *
                 ThermalFunctions::JbosonInterpolated(Ratio, 0) -
             2 * Temp * MassSquared *
-                ThermalFunctions::JbosonInterpolated(Ratio, 1));
+                ThermalFunctions::JbosonInterpolated(Ratio, 1)); //Missing term!!
```
vollous: Which missing term?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -3202,20 +3222,21 @@ double Class_Potential_Origin::V1Loop(const std::vector<double> &v,
     if (C_UseParwani)
     {
       for (std::size_t k = 0; k < NHiggs; k++)
-      {
-        res += HiggsMassesVec.at(k + NHiggs) *
-               boson(HiggsMassesVec[k], Temp, C_CWcbHiggs, 0);
-        res += boson(HiggsMassesVec[k], Temp, C_CWcbHiggs, -1);
+      { 
+        res += HiggsMassesVec.at(k + NHiggs) * bosonTderiv(HiggsMassesVec[k], HiggsMassesVec.at(k + NHiggs), Temp, C_CWcbHiggs, 1);
+        //std::cout << " --- debug " << HiggsMassesVec[k] << " " << HiggsMassesVec.at(k + NHiggs) << " " << HiggsMassesVec.at(k + NHiggs) * bosonTderiv(HiggsMassesVec[k], HiggsMassesVec.at(k + NHiggs), Temp, C_CWcbHiggs, 1) << " " << bosonTderiv(HiggsMassesVec[k], HiggsMassesVec.at(k + NHiggs), Temp, C_CWcbHiggs, -1) << std::endl; //debug Carlo
```
vollous: Please remove this comment

include/BSMPT/models/ClassPotentialOrigin.h
```diff
@@ -1088,7 +1088,8 @@ class Class_Potential_Origin
    *
    *
    */
-  double boson(double MassSquared, double Temp, double cb, int diff = 0) const;
+   double bosonTderiv(double MassSquared, double DevMassSquared, double Temp, double cb, int diff = 0) const;
+   double boson(double MassSquared, double Temp, double cb, int diff = 0) const;
```
vollous: Please fix the documentation here.

include/BSMPT/models/ClassPotentialOrigin.h
```diff
@@ -1088,7 +1088,8 @@ class Class_Potential_Origin
    *
    *
    */
-  double boson(double MassSquared, double Temp, double cb, int diff = 0) const;
+   double bosonTderiv(double MassSquared, double DevMassSquared, double Temp, double cb, int diff = 0) const;
```
vollous: This function might not be needed

vollous (changes requested):
ue to time constraints we hot-fixed the T-deriv with the PR that Christoph made.. Please review this PR and keep only the necessary changes for the field derivative corrections.

vollous:
Also, please manually increase the version number to `3.3.0`

src/models/ClassPotentialOrigin.cpp
rafaelboto: If Masses have a T dependence then there is also a piece (2 * Temp * MassSquared - Temp*Temp * DevMassSquared) 
The proposed way to handle is through bosonTderiv routine which has an additional argument

include/BSMPT/models/ClassPotentialOrigin.h
rafaelboto: Solution to T derivative of boson, with the additional argument. Is it possible to fix within boson?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -110,28 +110,29 @@ Class_Potential_Origin::CWTerm(double MassSquared, double cb, int diff) const
 double Class_Potential_Origin::boson(double MassSquared,
                                      double Temp,
                                      double cb,
-                                     int diff) const
+                                     int diff,
+                                     double DevMassSquared) const
 {
   double res = 0;
-  if (diff >= 0) res = CWTerm(std::abs(MassSquared), cb, diff);
+  if (diff >= 0) res = CWTerm(MassSquared, cb, diff);
```
phbasler, line 117: Why was the `std::abs` removed here?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -110,28 +110,29 @@ Class_Potential_Origin::CWTerm(double MassSquared, double cb, int diff) const
...
   if (diff == 0)
   {
     res += std::pow(Temp, 4) / (2 * std::pow(M_PI, 2)) *
            ThermalFunctions::JbosonInterpolated(Ratio);
   }
-  else if (diff == 1)
+  else if (diff > 0)
   {
     res += std::pow(Temp, 2) / (2 * std::pow(M_PI, 2)) *
-           ThermalFunctions::JbosonNumericalIntegration(Ratio, 1);
```
phbasler, line 127: Do you still need the `JbosonNumericalIntegration` function in the code?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -149,10 +150,11 @@ Class_Potential_Origin::fermion(double MassSquared, double Temp, int diff) const
     res += std::pow(Temp, 4) / (2 * std::pow(M_PI, 2)) *
            ThermalFunctions::JfermionInterpolated(Ratio);
   }
-  else if (diff == 1)
+
+  else if (diff > 0)
   {
     res += std::pow(Temp, 2) / (2 * std::pow(M_PI, 2)) *
-           ThermalFunctions::JfermionNumericalIntegration(Ratio, 1);
```
phbasler, line 155: `JfermionNumericalIntegration` can also be removed

src/models/ClassPotentialOrigin.cpp
```diff
@@ -3204,18 +3206,34 @@ double Class_Potential_Origin::V1Loop(const std::vector<double> &v,
       for (std::size_t k = 0; k < NHiggs; k++)
       {
         res += HiggsMassesVec.at(k + NHiggs) *
-               boson(HiggsMassesVec[k], Temp, C_CWcbHiggs, 0);
-        res += boson(HiggsMassesVec[k], Temp, C_CWcbHiggs, -1);
+               boson(HiggsMassesVec[k],
+                     Temp,
+                     C_CWcbHiggs,
+                     1,
```
phbasler: Why did the value here change from 0 to 1?

src/models/ClassPotentialOrigin.cpp
```diff
@@ -3204,18 +3206,34 @@ double Class_Potential_Origin::V1Loop(const std::vector<double> &v,
...
+        res += boson(GaugeMassesVec[k],
+                     Temp,
+                     C_CWcbGB,
+                     -1,
+                     GaugeMassesVec.at(k + NGauge));
       }
       for (std::size_t k = 0; k < NGauge; k++)
       {
-        res += 2 * boson(GaugeMassesZeroTempVec[k], Temp, C_CWcbGB, -1);
+        res += 2 * boson(GaugeMassesZeroTempVec[k], Temp, C_CWcbGB, -1, 0.0);
```
phbasler, line 3255: Why add `0.0`?  The function takes an int and so you have an type cast here. Same for the following calls.

src/models/ClassPotentialOrigin.cpp
```diff
@@ -3262,8 +3280,8 @@ double Class_Potential_Origin::V1Loop(const std::vector<double> &v,
         if (GaugeMassesVec[k] > 0)
         {
           VDebye += std::pow(GaugeMassesVec[k], 1.5);
-          VDebye += 1.5 * Temp * GaugeMassesVec.at(k + NGauge) *
-                    std::pow(GaugeMassesVec[k], 0.5);
+          VDebye += 1.5 * Temp * GaugeMassesVec[k + NGauge] *
```
phbasler, line 3317:
```suggestion
          VDebye += 1.5 * Temp * GaugeMassesVec.at(k + NGauge) *
```

Please use `at` instead of `[]` unless you have checked otherwise that `k+NGauge<GaugeMassesVec.size()` 
The `[]` operator will crash with a segfault if you are out of range

phbasler (changes requested):

src/models/ClassPotentialOrigin.cpp
vollous: When you take the `std::abs` value you lose information about the sign that you need for the derivatives. This was also checked numerically.

src/models/ClassPotentialOrigin.cpp
vollous: No, I'll remove it

src/models/ClassPotentialOrigin.cpp
vollous: I'll remove it

src/models/ClassPotentialOrigin.cpp
vollous: `boson` takes one more argument now, the mass derivative, which is that `0.0` . The `0.0` is not necessary but I kept for clarity

src/models/ClassPotentialOrigin.cpp
vollous: We still have the  `0` term below. This is an additional term that comes from the mass correction on the CW and $J_{B/F}$

src/models/ClassPotentialOrigin.cpp
vollous: I prefer to check the size before and use `[]`. I fear using `.at()` will make the code slower.

src/models/ClassPotentialOrigin.cpp
phbasler: Maybe we should have started it a long time ago but what do you think about adding this as comments here in the code?
Just a short comment marking the origin of these terms. This would increase readability in the long term and for new developers

src/models/ClassPotentialOrigin.cpp
phbasler: Where do we check that k + NGauge < GaugeMassesVec.size() ?

vollous: [Here](https://github.com/BSMPT/BSMPT/pull/299/changes#diff-9d2da88ab8960a426f8defe650f25adc04e9ca35f086183903fd9f3a1ab34e08R3058). This accounts for the case when `diff!=0`the mass vector have different lengths

```c++
// Length = N if diff = 0 else Length = 2N (mass_i, deriv_mass_i)
const size_t NMultiplier         = (diff == 0) ? 1 : 2;
const size_t NMultiplierFermions = (diff <= 0) ? 1 : 2;
if (HiggsMassesVec.size() != NHiggs * NMultiplier)
  throw("Missmatch NHiggs [V1Loop]");
if (GaugeMassesVec.size() != NGauge * NMultiplier)
  throw("Missmatch NGauge [V1Loop]");
if (GaugeMassesZeroTempVec.size() != NGauge * NMultiplier)
  throw("Missmatch NGauge T = 0 [V1Loop]");
if (QuarkMassesVec.size() != NQuarks * NMultiplierFermions)
  throw("Missmatch NQuarks [V1Loop]");
if (LeptonMassesVec.size() != NLepton * NMultiplierFermions)
  throw("Missmatch NLepton [V1Loop]");
(...)
if (HiggsMassesVec.size() != NHiggs * NMultiplier)
 throw("Missmatch NHiggs T = 0 [V1Loop]");
```

src/models/ClassPotentialOrigin.cpp
vollous: Sure, it actually helps a lot. I've added a few comments

phbasler (approved):

vollous (approved):

PR #301, opened 2026-05-11
Temperature derivative fix for Arnold-Espinosa and sound speed

### LINKED ISSUE #300: Missing terms in temperature derivative, slightly wrong derivative in sound speed calculation
The analytical derivative with respect to the temperature $T$ of the effective potential $V_{\text{eff}}$ is implemented incorrectly in the Arnold-Espinoa approach and missing some terms. It is used in the calculation of the phase transition strength parameter $\alpha$.

The problems are (all in `src/models/ClassPotentialOrigin.cpp`):
-  [L2466](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L2466): Inside function `Class_Potential_Origin::HiggsMassesSquared`, the Higgs mass matrix is called with the `diff` parameter,
  https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L2466 such that in case of the $T$-derivative, corresponding to `diff = -1`, only the $T$-derivative of the thermally corrected Higgs mass matrix is used and all other $T$-independent terms are dropped, which would cause e.g. all entries of the Higgs mass vector for zero temperature, `HiggsMassesZeroTempVec` in [L3143](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L3143), to be zero. However, `HiggsMassMatrix` here should always be called with `diff = 0` to return the full eigenvalues (with no derivative), since further down in the code e.g. in [L2527ff](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L2527), the $T$-derivative of the eigenvalues is then additionally calculated, which is what is actually needed.
- [L3248](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L3248) to [L3262](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L3262): In the calculation of the $T$-derivative of `VDebye`, the derivative of the thermally corrected mass eigenvalues of `HiggsMassesVec[k]` with respect to $T$ is not considered, therefore several terms are missing. With $V_{\text{daisy}}$ from e.g. https://arxiv.org/pdf/1803.02846, we have:
$\displaystyle\frac{\partial V_\text{daisy}}{\partial T} = -\frac{1}{12\pi}\sum_{\text{bosons}}\left[(\overline m^2)^{3/2} - (m^2)^{3/2}\right] - \frac32\frac{T}{12\pi}\sum_{\text{bosons}} (\overline m^2)^{1/2} \frac{\partial \overline m^2}{\partial T}$
and here the second term starting with the $\frac32$ factor is currently not implemented in the code.
- [L135](https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/models/ClassPotentialOrigin.cpp#L135): The numerical integration of the $T$-derivative of the bosonic $J$ function `ThermalFunctions::JbosonNumericalIntegration(Ratio, 1)` becomes highly numerically unstable for `Ratio < 0` and therefore should not be used. Instead, the $T$-derivative of the interplolated version, `ThermalFunctions::JbosonInterpolated(Ratio, 1)`, leads to more stable results.

Another different, but somewhat related issue was found during testing, showing that in the calculation of the sound speed is using a slightly wrong $T$-derivative, which can lead to instabilities in the calculation of the `alpha` parameter. Here, the problem is in `src/bounce_solution/bounce_solution.cpp` inside the function `BounceSolution::CalculateSoundSpeed`. The $T$-derivative of the effective potential is incorrectly applied also to the VEV, https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/bounce_solution/bounce_solution.cpp#L1173 and https://github.com/BSMPT/BSMPT/blob/47a174c9b1370558aaf94ded17b08a7c244032cc/src/bounce_solution/bounce_solution.cpp#L1175 when the VEV should stay fixed and the $T$-derivative should only be taken with respect to the $T$ argument of the effective potential.

cborschensky:
## Description

Fixes #300. The missing terms for the temperature derivative of the effective potential in the Arnold-Espinosa approach are added, and the interpolated $J$ functions, `ThermalFunctions::JfermionInterpolated` and `ThermalFunctions::JbosonInterpolated`, are now used in the $T$-derivative (for consistency both for the fermionic and the more problematic bosonic functions) instead of the numerical integrations `ThermalFunctions::JfermionNumericalIntegration` and `ThermalFunctions::JbosonNumericalIntegration`. Furthermore, the correct $T$-derivative is now used in the calculation of the sound speed (and switched partially to the implemented analytical $T$-derivative instead of a numerical derivative).

The numbers in the unit tests were updated, and an additional unit test is added to compare the analytical $T$-derivative of the effective potential with a simple numerical derivative, to ensure that in the $T$-derivative all terms are properly taken into account.

## Type of Change

 - [ ] Break (major)
 - [ ] Feature (minor)
 - [x] Bug fix (patch)


tests/unittests/Test-veffderivative.cpp
```diff
@@ -0,0 +1,103 @@
...
+                                              {0., 0., 100.},
+                                              {100., 200., 0.},
+                                              {50., 100., 200.}};
+
+  auto dTV_num = [=](std::vector<double> p, double T, double h)
+  {
+    double V1 = modelPointer->VEff(modelPointer->MinimizeOrderVEV(p), T - h, 0);
+    double V2 = modelPointer->VEff(modelPointer->MinimizeOrderVEV(p), T + h, 0);
+    return (V2 - V1) / 2. / h;
+  };
```
vollous: We have a numerical derivative implementation, it would be better to use it. You can use it like 
```c++
  const double eps = 1e-3;
  std::function<double(std::vector<double>, double)> dTV_num =
      [=](auto const &vev, auto const &T)
  {
    return NablaNumerical(
               {T},
               [&](std::vector<double> T)
               {
                 // Potential wrapper
                 return modelPointer->VEff(modelPointer->MinimizeOrderVEV(vev),
                                           T.at(0));
               },
               eps)
        .at(0);
  };
```

and call it 

```c++
double x_num = dTV_num_nabla(phi_val, T_val);
```
Its basically your funcion without the step size (if you want to can add it back in)

tests/unittests/Test-veffderivative.cpp
vollous, line 1: Move these tests to `tests/unittests/Test-origin.cpp`, its where we test the potential

tests/unittests/Test-veffderivative.cpp
```diff
@@ -0,0 +1,103 @@
...
+                                              {0., 0., 100., 0.},
+                                              {0., 50., 200., 0.},
+                                              {50., 100., 200., 10.}};
+
+  auto dTV_num = [=](std::vector<double> p, double T, double h)
+  {
+    double V1 = modelPointer->VEff(modelPointer->MinimizeOrderVEV(p), T - h, 0);
+    double V2 = modelPointer->VEff(modelPointer->MinimizeOrderVEV(p), T + h, 0);
+    return (V2 - V1) / 2. / h;
+  };
```
vollous: Do the same here please

vollous (changes requested):
We should use the built in numerical grad, i suggested an implementation

cborschensky:
> We should use the built in numerical grad, i suggested an implementation

I changed the numerical derivative to `NablaNumerical` both in the tests and in the sound speed calculation (where I take the numerical derivative of the analytical derivative, to get the second derivative wrt. $T$).

vollous (approved):
Looks Good. Thanks a lot!

